PR #6, opened 2024-02-09
tmf: Use query2D for querying histogram

arfio:
The previous implementation does a full query for each requested time. This patch now uses one query2D to calculate the same result.

[Change-Id: Ic84d8a46536718c097de65b4b17df927f9368755](https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/206051)


MatthewKhouzam:
This looks great. I would like to know, can we make a helper library with this that can be used for all xy plot data providers?

arfio:
I do not know if this specific change would be useful for all xy plot, but the DataDrivenXYProviderFactory can be used to create default XY data providers in a few lines of code that specify where to find the data in the state system.

bhufmann (approved):
LGTM. The bug reported initially on the previous Gerrit patch is fixed. Thanks.

MatthewKhouzam (approved):

PR #9, opened 2024-02-09
callstack.core: Merge callstack with profiling

arfio:
This patch moves everything from callstack to the profiling plugin.

[Change-Id: Ic3f109ef7fe2c88467a29913f41264ed5f3edc6d](https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/205232)


arfio:
> Patrick Tasse - Feb 05
> I get an Unnecessary SuppressWarning error here?
I removed it

> Patrick Tasse - Feb 05
> The [o.e.tc](http://o.e.tc/).analysis.callstack.doc.user might also need to be renamed/moved. It can be done in a separate patch. There might be some rewording necessary in the documentation.
Yes I agree, should I add this change to this pull request ?

PatrickTasse:
> Yes I agree, should I add this change to this pull request ?

Sure!

PatrickTasse (approved):

PR #15, opened 2024-02-12
tmf: Handle multiple jobs when updating statistics

arfio:
When one job is already scheduled, the previous logic did not create another job preventing subsequent selection from updating the UI in the statistics view. This commit always computes a new job and cancels the previous one if present.

[Fixed] Updating statistics now handles concurrent jobs correctly

[Change-Id: I14c99c6eac40def82338ca9341031951a4703db4](https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/206052)


bhufmann (changes requested):
Code looks good to me. The Gerrit change-id is not required anymore after the migration to GitHub. Could you please remove it?

bhufmann (approved):
LGTM. Thanks.

MatthewKhouzam (approved):
Looks great!

Issue #18, opened 2024-02-16
TimeRangeException for experiments in Event Density view


bhufmann:
With commit: 15d6f25 I get the following exception for the `Event Density` view when opening an experiment:

```
Exception in thread "Line chart update" org.eclipse.tracecompass.statesystem.core.exceptions.TimeRangeException: Time conditions 1487359244565151799 is lower than state system start time: 1487359244680159245
	at org.eclipse.tracecompass.internal.statesystem.core.StateSystem.query2D(StateSystem.java:652)
	at org.eclipse.tracecompass.internal.statesystem.core.StateSystem.query2D(StateSystem.java:634)
	at org.eclipse.tracecompass.tmf.core.statistics.TmfStateStatistics.histogramQuery(TmfStateStatistics.java:129)
	at org.eclipse.tracecompass.internal.tmf.core.histogram.HistogramDataProvider.fetchXY(HistogramDataProvider.java:128)
	at org.eclipse.tracecompass.tmf.core.model.xy.TmfTreeXYCompositeDataProvider.getXyResponses(TmfTreeXYCompositeDataProvider.java:187)
	at org.eclipse.tracecompass.tmf.core.model.xy.TmfTreeXYCompositeDataProvider.fetchXY(TmfTreeXYCompositeDataProvider.java:136)
	at org.eclipse.tracecompass.tmf.ui.viewers.xychart.linechart.TmfCommonXAxisChartViewer$UpdateThread.updateData(TmfCommonXAxisChartViewer.java:384)
	at org.eclipse.tracecompass.tmf.ui.viewers.xychart.linechart.TmfCommonXAxisChartViewer$UpdateThread.run(TmfCommonXAxisChartViewer.java:350)
```

bhufmann:
@arfio could this exception be caused by the change #6?

arfio:
Yes it was, I was sure I fixed this issue but cannot find where, I must have removed the fix at some point.

bhufmann:
Fixed with above pull request.

PR #19, opened 2024-02-17
tmf: Fix for times requested before start time

arfio:
When requested times are below the start time of the state system when doing a query2d, it throws an error. This patch fixes that issue by replace with the start time when doing the query2d.

[Fixes] #18

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
@@ -119,11 +119,14 @@ public class TmfStateStatistics implements ITmfStatistics {
         }
         List<Long> times = new ArrayList<>();
         for (int i = 0; i < timeRequested.length; i++) {
-            if (timeRequested[i] > fTotalsStats.getCurrentEndTime()) {
+            if (timeRequested[i] < fTotalsStats.getStartTime()) {
+                times.add(fTotalsStats.getStartTime());
+            } else if (timeRequested[i] < fTotalsStats.getCurrentEndTime()) {
+                times.add(timeRequested[i]);
+            } else {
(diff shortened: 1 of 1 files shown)


bhufmann (approved):
LGTM. Thanks for the quick response!

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
```diff
@@ -119,11 +119,14 @@ public void dispose() {
         }
         List<Long> times = new ArrayList<>();
         for (int i = 0; i < timeRequested.length; i++) {
-            if (timeRequested[i] > fTotalsStats.getCurrentEndTime()) {
+            if (timeRequested[i] < fTotalsStats.getStartTime()) {
+                times.add(fTotalsStats.getStartTime());
```
PatrickTasse, line 123: It should ensure that the start time is added only once. For end time it's only once because it then immediately breaks out of the loop.

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
arfio: The issue is that the result should have as many times between the given array and the resulting list. I tried to fill the resulting list with 0 and do a query2d only for one start and the rest of the times but it was not working correctly. Although, I am not sure why.

MatthewKhouzam: @PatrickTasse @arfio can we resolve this?

PatrickTasse (approved):
Let's fix the exception and worry later if it's slightly inefficient :)

PR #20, opened 2024-02-20
ctf: support out of spec index files from LTTng

arfio:
IDX files were added to help speed up trace indexing. This patch will load them. This patch should not affect the system performance, but might lay down some groundwork to improve streaming performance.

The index format is not yet part of the CTF standard, but is described here:
https://git.efficios.com/?p=babeltrace.git;a=blob;f=src/plugins/ctf/fs-src/lttng-index.hpp;h=36dc7cc934d46366aa12ebf7b13f239e8699eacb;hb=HEAD

### DIFF
--- a/ctf/org.eclipse.tracecompass.ctf.core/src/org/eclipse/tracecompass/ctf/core/trace/CTFStreamInput.java
+++ b/ctf/org.eclipse.tracecompass.ctf.core/src/org/eclipse/tracecompass/ctf/core/trace/CTFStreamInput.java
@@ -108,14 +108,13 @@ public class CTFStreamInput implements IDefinitionScope {
      *            The stream to which this StreamInput belongs to.
      * @param file
      *            Information about the trace file (for debugging purposes).
-     * @since 2.0
+     * @since 4.5
      */
-    public CTFStreamInput(ICTFStream stream, File file) {
+    public CTFStreamInput(ICTFStream stream, File file, StreamInputPacketIndex index) {
         fStream = stream;
         fFile = file;
         fFileName = fFile.getName();
-
-        fIndex = new StreamInputPacketIndex();
+        fIndex = index;
         /*
          * Create the definitions we need to read the packet headers + contexts
          */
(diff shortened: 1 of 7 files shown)


bhufmann (changes requested):
Multiple tests are failing (see GitHub actions output) with errors that are probably related to this change. Could you please look into this? Thanks

arfio:
Writing traces is failing because the context header struct definition is missing when reading the index files. This struct definition is needed when trimming the traces among other use cases. Is there any way of reconstructing that definition elsewhere, or should we just use the index files to calculate the end time and not store it for now ?

bhufmann:
> Writing traces is failing because the context header struct definition is missing when reading the index files. This struct definition is needed when trimming the traces among other use cases. Is there any way of reconstructing that definition elsewhere, or should we just use the index files to calculate the end time and not store it for now ?

@MatthewKhouzam  any suggestions?

MatthewKhouzam:
We have discussed this offline, here is the results:

1. Read the index files. You only care about the packet starts.
2. Read the metadata
3. Quickly populate with any known index locations from the CTF stream

This should fix any issue.

MatthewKhouzam (dismissed):
I wrote this, I feel odd approving it.

MatthewKhouzam (commented):
TraceReadAllTracesTest.readTraces:108 DJANGO_CLIENT Trace read error k_0
  TraceReadAllTracesTest.readTraces:108 DJANGO_DB Trace read error k_0
  TraceReadAllTracesTest.readTraces:108 DJANGO_HTTPD Trace read error k_0

arfio:
> TraceReadAllTracesTest.readTraces:108 DJANGO_CLIENT Trace read error k_0 TraceReadAllTracesTest.readTraces:108 DJANGO_DB Trace read error k_0 TraceReadAllTracesTest.readTraces:108 DJANGO_HTTPD Trace read error k_0

This is due to a mismatch in index file version, I now throw if the minor is different, and I skip reading the index file.

MatthewKhouzam (approved):

bhufmann (approved):
The errors are gone. I opened a couple of LTTng CTF traces and non-LTTng CTF traces and they both work. Thanks for the contribution.

PR #23, opened 2024-02-22
profiling.core: make callstack api

arfio:
This commit makes the new callstack api, moving necessary classes to provisional packages and adding it to the API.


MatthewKhouzam (approved):

MatthewKhouzam (approved):
Looks good to me.

analysis/org.eclipse.tracecompass.analysis.profiling.core/META-INF/MANIFEST.MF
```diff
@@ -42,12 +42,12 @@ Export-Package: org.eclipse.tracecompass.analysis.profiling.core.base;uses:="org
...
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.base;x-friends:="org.eclipse.tracecompass.analysis.profiling.core.tests,org.eclipse.tracecompass.analysis.profiling.ui",
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.callgraph;x-friends:="org.eclipse.tracecompass.analysis.profiling.core.tests",
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.callstack;x-friends:="org.eclipse.tracecompass.analysis.profiling.ui,org.eclipse.tracecompass.analysis.profiling.core.tests",
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.instrumented;x-friends:="org.eclipse.tracecompass.analysis.profiling.core.tests",
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.model;x-friends:="org.eclipse.tracecompass.lttng2.ust.core,org.eclipse.tracecompass.analysis.profiling.core.tests",
- org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.tree;x-friends:="org.eclipse.tracecompass.analysis.profiling.ui,org.eclipse.tracecompass.analysis.profiling.core.tests"
+ org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.base,
```
bhufmann, line 45: provisional APIs are APIs that are proposed and are supposed to be made public in the future. However, they still need to be marked as internal in the MANIFEST.MF (x-internal:=true or x-friends if there are any). If this PR is to make the APIs public, then the package name cannot contain provisional.

This applies for all relevant changed lines in the MANIFEST.MF files.

bhufmann (changes requested):
Beside the main comment below, I noticed that there are API that were already in the mainline callstack packages that were also in the incubator callstack packages. I think Genevieve added this classes/interface to make the mainlining easier. For example, if you compare the following 2 packages you find duplicate classes:
`org.eclipse.tracecompass.analysis.profiling.core.callstack`  
`org.eclipse.tracecompass.internal.provisional.analysis.profiling.core.callstack'

Also, the interface `IFlameChartProvider` exists twice in the same plug-in.

What is your plan handling these duplications? Please let me know.

arfio:
Once this path is merged, we can deprecate the old versions of the duplicated classes as well as the old callstack API. Then, after some time has passed, we will be able to remove the old API and replace it completely.

bhufmann (approved):
Thanks for this contribution. It looks good to me.  You can merge it, since you're a committer now.

PR #24, opened 2024-02-22
callstack.core: Add arrows to flamechart test

arfio:
This patch adds arrow in the CallStackProviderStub to test the data provider. It also adds checks for arrows in the testFetchModel of the FlameChartDataProviderTest


MatthewKhouzam (approved):

arfio:
Previous comments on https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/204666 were addressed 
But this comment was not:

> Also, should it return a link that crosses the range but does not start or end within the range?

In some cases, it should, in others, links add so much clutter that it should not. So I am not sure about including those links.

MatthewKhouzam:
@PatrickTasse Any chance to get a second review?

bhufmann (approved):
Looks good to me, Thanks!

PR #26, opened 2024-02-27
tmf: Fix statistics queries when there is no intervals

arfio:
There are still two issues with the previous implementation:

- Having the duplicates of the start time in the TimeArrayCondition will slow down the execution of the binary search.
- When the query2d returned no intervals, an exception is thrown. This situation was never encountered because the previous version would add at least the end time or the start time but that would lead to an incorrect first bucket or last bucket

This patch was the one I lost. I modified it a bit to reduce the complexity of the queryHistogram method.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
@@ -117,40 +117,63 @@ public class TmfStateStatistics implements ITmfStatistics {
         if (quark == ITmfStateSystem.INVALID_ATTRIBUTE) {
             return Collections.emptyList();
         }
-        List<Long> times = new ArrayList<>();
+        List<@NonNull Long> times = new ArrayList<>();
         for (int i = 0; i < timeRequested.length; i++) {
+            /* create padding for every window of time before the start time */
             if (timeRequested[i] < fTotalsStats.getStartTime()) {
-                times.add(fTotalsStats.getStartTime());
-            } else if (timeRequested[i] < fTotalsStats.getCurrentEndTime()) {
-                times.add(timeRequested[i]);
+                list.add(0L);
             } else {
+                if (!list.isEmpty()) {
+                    /* replace this bucket with the [start time, next time] */
+                    times.add(fTotalsStats.getStartTime());
+                    list.remove(list.size() - 1);
+                }
+                times.add(timeRequested[i]);
+            }
+            if (timeRequested[i] > fTotalsStats.getCurrentEndTime()) {
                 times.add(fTotalsStats.getCurrentEndTime());
                 break;
             }
         }
         try (FlowScopeLog log = new FlowScopeLogBuilder(LOGGER, Level.FINE, "StateStatistics:histogramQuery").build()) { //$NON-NLS-1$
+            addHistogramValuesFromQuery2d(quark, times, list);
+            /* Padding for after trace ends */
+            for (int i = list.size(); i < timeRequested.length; i++) {
+                list.add(0L);
+            }
+            return list;
+        }
+    }
+
+    private List<@NonNull Long> addHistogramValuesFromQuery2d(int quark, List<@NonNull Long> times, List<@NonNull Long> results) {
+        try {
             Iterable<@NonNull ITmfStateInterval> intervals = fTotalsStats.query2D(Collections.singletonList(quark), times);
             List<@NonNull ITmfStateInterval> sortedIntervals = Lists.newArrayList(intervals);
+            if (sortedIntervals.isEmpty()) {
+                return results;
+            }
             sortedIntervals.sort(Comparator.comparingLong(ITmfStateInterval::getStartTime));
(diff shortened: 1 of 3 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
```diff
@@ -117,40 +117,63 @@ public void dispose() {
...
             List<@NonNull ITmfStateInterval> sortedIntervals = Lists.newArrayList(intervals);
+            if (sortedIntervals.isEmpty()) {
+                return results;
+            }
             sortedIntervals.sort(Comparator.comparingLong(ITmfStateInterval::getStartTime));
-            int j = 0;
             long previousTotal;
-            if (!sortedIntervals.isEmpty() && fTotalsStats.getStartTime() != sortedIntervals.get(0).getStartTime()) {
+            if (fTotalsStats.getStartTime() != sortedIntervals.get(0).getStartTime()) {
                 previousTotal = sortedIntervals.get(0).getValueInt();
```
MatthewKhouzam, line 158: getValueLong? asking for a friend?

MatthewKhouzam (commented):
Quick comment. I approve otherwise.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
arfio: Oops, fixed! Thanks

MatthewKhouzam (dismissed):

MatthewKhouzam:
Can you get the build to pass, then I will approve?

MatthewKhouzam (dismissed):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
```diff
@@ -117,40 +117,63 @@ public void dispose() {
...
+                }
+                times.add(timeRequested[i]);
+            }
+            if (timeRequested[i] > fTotalsStats.getCurrentEndTime()) {
                 times.add(fTotalsStats.getCurrentEndTime());
                 break;
             }
         }
         try (FlowScopeLog log = new FlowScopeLogBuilder(LOGGER, Level.FINE, "StateStatistics:histogramQuery").build()) { //$NON-NLS-1$
+            addHistogramValuesFromQuery2d(quark, times, list);
```
bhufmann, line 139: Opening the event densitiv view in Eclipse or when running the trace server, this method throws a ClassCastException which is swallowed. I added a try- catch to get the stack back trace. For that reason the histogram/Overview view is not populated with theia-trace-extension/vscode-trace-extension and event density view in Eclipse.
```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
	at org.eclipse.tracecompass.statesystem.core.interval.ITmfStateInterval.getValueLong(ITmfStateInterval.java:110)
	at org.eclipse.tracecompass.tmf.core.statistics.TmfStateStatistics.addHistogramValuesFromQuery2d(TmfStateStatistics.java:171)
	at org.eclipse.tracecompass.tmf.core.statistics.TmfStateStatistics.histogramQuery(TmfStateStatistics.java:140)
	at org.eclipse.tracecompass.internal.tmf.core.histogram.HistogramDataProvider.fetchXY(HistogramDataProvider.java:128)
	at org.eclipse.tracecompass.tmf.core.model.xy.TmfTreeXYCompositeDataProvider.getXyResponses(TmfTreeXYCompositeDataProvider.java:187)
	at org.eclipse.tracecompass.tmf.core.model.xy.TmfTreeXYCompositeDataProvider.fetchXY(TmfTreeXYCompositeDataProvider.java:136)
	at org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.DataProviderService.getXY(DataProviderService.java:390)
```

bhufmann (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/statistics/TmfStateStatistics.java
arfio: I modified the Statistics analysis to generate long state values instead of integers, so all statistics ht files will have to be regenerated. I did not observe this when I regenerate the ht files.

bhufmann: You need to update the version number of the state provider of the statistics then. Otherwise, the state systems are not rebuilt for existing traces and users will get that issue.

MatthewKhouzam (approved):

bhufmann (approved):
Looks good to me. Thanks!

PR #29, opened 2024-02-29
callstack.core: Add icon to new callstack analysis

bhufmann:


MatthewKhouzam (approved):
Thanks!

PR #33, opened 2024-03-11
Remove Eclipse dependencies in ctf.core plugin

bhufmann:
- ctf.core: Remove direct Eclipse platform dependency from ctf.core
- ctf.core: Remove dependency to o.e.tc.common.core

By removing this dependency this plug-in will be independent to other Eclipse plug-ins and can be deployed as maven artifact with less third-party dependencies.

Note, that class SaturatedArithmetic is copied from plug-in org.eclipse.tracecompass.common.core. The original tests are refactored and a test suite was added in ctf.core.tests added so that the new class is tested, too.


MatthewKhouzam (approved):
Looks good to me! Thanks!

PR #41, opened 2024-03-19
releng: Build with tracecompass-e4.31 target by default

PatrickTasse:


MatthewKhouzam (dismissed):
Don't need the changeid

rcp/org.eclipse.tracecompass.rcp/legacy-e4.30/feature.xml
```diff
@@ -0,0 +1,676 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature
+      id="org.eclipse.tracecompass.rcp"
+      label="%featureName"
+      version="9.2.0.qualifier"
```
bhufmann, line 5: shouldn't it be 9.3.0?

bhufmann (changes requested):

bhufmann (approved):

PR #45, opened 2024-03-20
statesysten: add processSpanStateChange function to statesystem

farajidaneshgar:
* A new function (processSpanStateChange) is added to the TransientState class. Because in async traces unlike sync traces, the position of  the event in fOngoingInfo is not the same as the quark of the event in the call stack. So we should pass two integer, quark and callstack quark to the function.

* The Statesystem and ItmfStateSystem classes are changed relayed to this function.
this changes is for adding flamgraph analysis for opentracing

### DIFF
--- a/statesystem/org.eclipse.tracecompass.statesystem.core/src/org/eclipse/tracecompass/internal/statesystem/core/StateSystem.java
+++ b/statesystem/org.eclipse.tracecompass.statesystem.core/src/org/eclipse/tracecompass/internal/statesystem/core/StateSystem.java
@@ -681,4 +681,19 @@ public class StateSystem implements ITmfStateSystemBuilder {
                 attribute + " at time " + timestamp + //$NON-NLS-1$
                 ", returning dummy interval"); //$NON-NLS-1$
     }
+
+    /**
+     * @param t
+     * @param value
+     * @param attributeQuark
+     * @param spanQuark
+     * @throws TimeRangeException
+     * @throws StateValueTypeException
+     */
+    @Override
+    public void modifySpanAttribute(long t, Object value, int attributeQuark, int spanQuark)
+            throws TimeRangeException, StateValueTypeException {
+        transState.processSpanStateChange(t, value, attributeQuark,spanQuark);
+    }
+
 }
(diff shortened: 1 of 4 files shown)


statesystem/org.eclipse.tracecompass.statesystem.core/src/org/eclipse/tracecompass/internal/statesystem/core/StateSystem.java
```diff
@@ -681,4 +681,19 @@ static void logMissingInterval(int attribute, long timestamp) {
                 attribute + " at time " + timestamp + //$NON-NLS-1$
                 ", returning dummy interval"); //$NON-NLS-1$
     }
+
+    /**
+     * @param t
```
MatthewKhouzam, line 686: More doc please

statesystem/org.eclipse.tracecompass.statesystem.core/src/org/eclipse/tracecompass/internal/statesystem/core/StateSystem.java
```diff
@@ -681,4 +681,19 @@ static void logMissingInterval(int attribute, long timestamp) {
...
+     * @param value
+     * @param attributeQuark
+     * @param spanQuark
+     * @throws TimeRangeException
+     * @throws StateValueTypeException
+     */
+    @Override
+    public void modifySpanAttribute(long t, Object value, int attributeQuark, int spanQuark)
```
MatthewKhouzam, line 694: Span is not a state system notion, it's a contextual interval. Should we bring it down to this level of abstraction?

statesystem/org.eclipse.tracecompass.statesystem.core/src/org/eclipse/tracecompass/statesystem/core/ITmfStateSystem.java
```diff
@@ -540,4 +541,15 @@ int getQuarkRelative(int startingNodeQuark, String... subPath)
      */
     Iterable<@NonNull ITmfStateInterval> query2D(@NonNull Collection<Integer> quarks,
             long start, long end) throws StateSystemDisposedException, IndexOutOfBoundsException, TimeRangeException;
+
+    /**
+     * @param t
```
MatthewKhouzam, line 546: More documentation

MatthewKhouzam (changes requested):
Transient State is an implementation of the API. Are we certain we want to add an API for this feature? It seems... like everything else will need to implement "unsupported" though it's designed for one implementation.

Issue #47, opened 2024-03-20
[TMF] Entering a window span of 1ns in Histogram View should be invalid


vladarama:
Original Issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=550946

PR #48, opened 2024-03-20
tmf: fix invalid 1ns histogram span

vladarama:
Fixes: #47 

This PR fixes an issue where histograms would be allowed to have a 1ns window span, which should not be possible.

Signed-off-by: Vlad Arama <vlad.arama@ericsson.com>

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
+++ b/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
@@ -610,6 +610,11 @@ public class HistogramView extends TmfView implements ITmfTimeAligned {
             realStart = fTraceStartTime;
         }
 
+        if (duration == 1) {
+            fTimeSpanControl.setValue(fWindowSpan);
+            return;
+        }
+        
         long endTime = realStart + duration;
         if (endTime > fTraceEndTime) {
             endTime = fTraceEndTime;
--- a/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
+++ b/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
@@ -610,7 +610,7 @@ public class HistogramView extends TmfView implements ITmfTimeAligned {
             realStart = fTraceStartTime;
         }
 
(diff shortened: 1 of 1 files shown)


tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
```diff
@@ -610,6 +610,11 @@ private void setNewRange(long startTime, long duration) {
             realStart = fTraceStartTime;
         }
 
+        if (duration == 1) {
+            fTimeSpanControl.setValue(fWindowSpan);
```
MatthewKhouzam, line 614: Do we need this?

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
```diff
@@ -610,6 +610,11 @@ private void setNewRange(long startTime, long duration) {
             realStart = fTraceStartTime;
         }
 
+        if (duration == 1) {
```
MatthewKhouzam, line 613: let's do <= 1? that way we can catch 0 too?

MatthewKhouzam (changes requested):
Test with negative time ranges for giggles?

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
vladarama: Good idea 👍

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/views/histogram/HistogramView.java
vladarama: This sets the value in the text box to its previous value. 
I.e. if we don't use that, we will still have 1ns still written in the text box even if the histogram isn't updated.

MatthewKhouzam: Indeed, I misread duration... oops :)

MatthewKhouzam (approved):
Good w/ me!

bhufmann (approved):
LGTM. Thanks!

PR #49, opened 2024-03-27
tmf: Add utility to detect directory traces from trace path

bhufmann:
This utility checks whether the parent or grandparent path of given path to a file is a valid directory trace. If it is a directory trace then return the parent or grandparent path.

[Added] TmfTraceType.checkAndUpdateTracePath() to check directory traces


marcdumais-work (approved):
(symbolic approval)
LTGM!

PatrickTasse (approved):

PR #50, opened 2024-04-01
tmf.ui: add support for unicode character marker symbols

christophebedard:
This is very useful to be able to identify markers by letters or more importantly by numbers to show sequences.

Successor to https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/128504

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/StyleProperties.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/StyleProperties.java
@@ -283,7 +283,10 @@ public final class StyleProperties {
     /**
      * Symbol type as a string. Possible values: {@link SymbolType}.
      * <p>
-     * Default: @link {@link SymbolType#NONE}
+     * Default: {@link SymbolType#NONE}
+     * <p>
+     * If the value is not amongst {@link SymbolType}, the string value itself
+     * is drawn.
      */
     public static final String SYMBOL_TYPE = "symbol-type"; //$NON-NLS-1$
 
--- a/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
+++ b/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
@@ -2314,7 +2314,8 @@ public class TimeGraphControl extends TimeGraphBaseControl
             } else if (VerticalAlign.TOP.equals(verticalAlign)) {
                 y -= Math.max(0, rect.height / 2 - symbolSize);
             }
-            switch (String.valueOf(symbolType)) {
+            String symbol = String.valueOf(symbolType);
+            switch (symbol) {
             case SymbolType.CROSS:
                 SymbolHelper.drawCross(gc, color, symbolSize, rect.x, y);
                 break;
@@ -2333,8 +2334,22 @@ public class TimeGraphControl extends TimeGraphBaseControl
             case SymbolType.CIRCLE:
                 SymbolHelper.drawCircle(gc, color, symbolSize, rect.x, y);
                 break;
-            default:
+            case SymbolType.DIAMOND:
                 SymbolHelper.drawDiamond(gc, color, symbolSize, rect.x, y);
+                break;
+            default:
+                Color oldColor = gc.getForeground();
+                gc.setForeground(color);
+                int height = (int) (rect.height * heightFactor);
+                TimeGraphRender.setFontForHeight(height, gc);
+                int textSize = (gc.textExtent(symbol).y + 1) / 2;
+                if (VerticalAlign.BOTTOM.equals(verticalAlign)) {
+                    y = rect.y + rect.height / 2 + Math.max(0, rect.height / 2 - textSize);
+                } else if (VerticalAlign.TOP.equals(verticalAlign)) {
+                    y = rect.y + rect.height / 2 - Math.max(0, rect.height / 2 - textSize);
+                }
+                gc.drawText(symbol, rect.x - symbolSize, y - textSize, true);
+                gc.setForeground(oldColor);
             }
             gc.setAlpha(OPAQUE);
             if (marker.getDuration() == 0) {


tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -155,6 +155,17 @@ public class TimeGraphControl extends TimeGraphBaseControl
...
+    /**
+     * Lowest unicode value for an emoji:
+     * https://unicode.org/Public/emoji/15.0/emoji-sequences.txt
+     */
+    private static final int EMOJI_LOW = 0x200D;
+    /**
+     * Highest unicode value for an emoji:
+     * https://unicode.org/Public/emoji/15.0/emoji-sequences.txt
+     */
+    private static final int EMOJI_HIGH = 0x1f9d1;
```
PatrickTasse, line 167: Looking at the link provided, there are now emojis up to 1faf8?

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -2333,8 +2345,21 @@ protected void drawMarker(IMarkerEvent marker, Rectangle bounds, ITimeDataProvid
...
                 SymbolHelper.drawDiamond(gc, color, symbolSize, rect.x, y);
+                break;
+            default:
+                char[] ca = symbol.toCharArray();
+                int first = ca[0];
+                /* check if the character is of length 1 or an emoji */
+                if (ca.length == 1 || (first >= EMOJI_LOW && first <= EMOJI_HIGH)) {
+                    Color oldColor = gc.getForeground();
+                    gc.setForeground(color);
+                    gc.drawText(symbol, rect.x, rect.y, true);
```
PatrickTasse, line 2358: The size of the emoji should be used to set the font:
TimeGraphRender.setFontForHeight(symbolSize * 2, gc);

The coordinates should be adjusted so that (rect.x, y) is the center of the emoji:
gc.drawText(symbol, rect.x - symbolSize, y - symbolSize, true);
This assumes that the emoji width is the same as its height.
It could be an ok assumption for a single character, otherwise we would need to calculate with gc.stringExtent (this is slow).

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -2314,7 +2325,8 @@ protected void drawMarker(IMarkerEvent marker, Rectangle bounds, ITimeDataProvid
             } else if (VerticalAlign.TOP.equals(verticalAlign)) {
                 y -= Math.max(0, rect.height / 2 - symbolSize);
             }
-            switch (String.valueOf(symbolType)) {
+            String symbol = String.valueOf(symbolType);
```
PatrickTasse, line 2328: The Javadoc of StyleProperties.SYMBOL_TYPE should be updated.

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
christophebedard: > TimeGraphRender.setFontForHeight(symbolSize * 2, gc);

I used `(int) (rect.height * heightFactor)` directly instead of `symbolSize * 2` (which is `2 * (int) Math.ceil(rect.height * heightFactor / 2)`), otherwise we may get a slightly wrong value due to the conversion to `int` with `(int) Math.ceil(double)`.

> This assumes that the emoji width is the same as its height.
It could be an ok assumption for a single character, otherwise we would need to calculate with gc.stringExtent (this is slow).

Yeah, I think it's fine to try to horizontally center on the first emoji.

----

One thing I noticed is that an emoji seems to be drawn slightly bigger (specifically taller) than a normal character (same height factor of `1.0f`):

image

but I believe that it is pretty normal and expected: A📤. It can be adjusted using the height factor anyway.

christophebedard (commented):
Thanks for the review!

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -2333,8 +2345,23 @@ protected void drawMarker(IMarkerEvent marker, Rectangle bounds, ITimeDataProvid
...
                 break;
-            default:
+            case SymbolType.DIAMOND:
                 SymbolHelper.drawDiamond(gc, color, symbolSize, rect.x, y);
+                break;
+            default:
+                char[] ca = symbol.toCharArray();
+                int first = ca[0];
+                /* check if the string is of length 1 or if the first character is an emoji */
+                if (ca.length == 1 || (first >= EMOJI_LOW && first <= EMOJI_HIGH)) {
```
christophebedard, line 2355: I looked at this again. This limits what can be drawn to single letters or numbers _or_ any number of emojis. What if we want to draw a normal double- or triple-digit number? What about just checking if the symbol type value is (not) `SymbolType.NONE`:

```suggestion
                /* draw symbol type value directly if not SymbolType.NONE */
                if (!symbol.equals(SymbolType.NONE)) {
```

PatrickTasse: > One thing I noticed is that an emoji seems to be drawn slightly bigger (specifically taller) than a normal character (same height factor of 1.0f):

I think that's because the emoji takes the full height given, but fonts keep some height for ascenders and descenders?

> I looked at this again. This limits what can be drawn to single letters or numbers or any number of emojis. What if we want to draw a normal double- or triple-digit number? What about just checking if the symbol type value is (not) SymbolType.NONE:

I agree, you can have any length string output as long as the first character is an emoji now. So I think it's ok to accept any length string as the symbol type (except for the reserved values e.g. "square"...).

Then there would be no case for the DIAMOND default. Please update the code and Javadoc accordingly.

christophebedard: > I think that's because the emoji takes the full height given, but fonts keep some height for ascenders and descenders?

That makes sense.

> I agree, you can have any length string output as long as the first character is an emoji now. So I think it's ok to accept any length string as the symbol type (except for the reserved values e.g. "square"...).
> 
> Then there would be no case for the DIAMOND default. Please update the code and Javadoc accordingly.

Done. Let me know if that's what you had in mind.

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -2333,8 +2334,17 @@ protected void drawMarker(IMarkerEvent marker, Rectangle bounds, ITimeDataProvid
             case SymbolType.CIRCLE:
                 SymbolHelper.drawCircle(gc, color, symbolSize, rect.x, y);
                 break;
-            default:
+            case SymbolType.DIAMOND: // Fall-through
+            case SymbolType.NONE:
```
PatrickTasse, line 2338: NONE cannot really happen here because it is checked in the containing if-clause.
If we must put it explicitly (perhaps for static code quality checkers?), it would be clearer to separate it with just a break to make it clear that it does not draw anything.

In fact, for NONE it will draw the name of the annotation (from the annotation model, not the style)...

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/StyleProperties.java
```diff
@@ -283,7 +283,10 @@ public interface FontWeight {
     /**
      * Symbol type as a string. Possible values: {@link SymbolType}.
      * <p>
-     * Default: @link {@link SymbolType#NONE}
+     * Default: {@link SymbolType#NONE}, which is {@link SymbolType#DIAMOND}.
```
PatrickTasse, line 286: Revert this change, since it does not actually draw a diamond for NONE.

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
christophebedard: oh I completely missed that. No, it doesn't complain about a missing `case SymbolType.NONE`, so I just removed it.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/StyleProperties.java
christophebedard: done

christophebedard:
I removed "single" from the PR title (like I did for the commit message), since it doesn't only show _single_ unicode characters anymore.

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
```diff
@@ -2333,8 +2334,16 @@ protected void drawMarker(IMarkerEvent marker, Rectangle bounds, ITimeDataProvid
...
-            default:
+            case SymbolType.DIAMOND:
                 SymbolHelper.drawDiamond(gc, color, symbolSize, rect.x, y);
+                break;
+            default:
+                Color oldColor = gc.getForeground();
+                gc.setForeground(color);
+                int height = (int) (rect.height * heightFactor);
+                TimeGraphRender.setFontForHeight(height, gc);
+                gc.drawText(symbol, rect.x - symbolSize, y - symbolSize, true);
```
PatrickTasse, line 2345: For some reason, the height of the text bounds is not actually what was requested [*], so its half-height is not equal to symbolSize. Unfortunately we must compute the text height and use it not only to adjust the top of text bounds relative to the requested middle (y) but we must also recalculate the middle y for TOP and BOTTOM alignment.

This works for me:

                int textSize = (gc.textExtent(symbol).y + 1) / 2;
                if (VerticalAlign.BOTTOM.equals(verticalAlign)) {
                    y = rect.y + rect.height / 2 + Math.max(0, rect.height / 2 - textSize);
                } else if (VerticalAlign.TOP.equals(verticalAlign)) {
                    y = rect.y + rect.height / 2 - Math.max(0, rect.height / 2 - textSize);
                }
                gc.drawText(symbol, rect.x - symbolSize, y - textSize, true);

[*] In my debugging, when we request a font with height of 100 pixels, this corresponds to a requested font data height of 75 points (with DPI=96 px/in and PPI=72 pts/in), but the created font data's height is 113 (apparently in pixels although the API says this should be in points), and the displayed text indeed is 113 pixels in height.

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/widgets/timegraph/widgets/TimeGraphControl.java
christophebedard: Oh yeah, it is much more apparent when zooming in.

Does this only replace `gc.drawText()`, or does it also replace `TimeGraphRender.setFontForHeight()`? I don't think `TimeGraphRender.setFontForHeight()` is actually doing anything.

PatrickTasse: The code should replace only line 2345. TimeGraphRender.setFontForHeight() will set the font on the gc for the following drawText operation(s). It will either create (first time) or reuse a font with the expected height.

christophebedard: Sounds good, that's what I did :grin:

PatrickTasse (approved):

PR #51, opened 2024-04-02
profiling.core: add weighted tree tests from incubator

arfio:
Move weighted tree tests from incubator that were forgotten when porting callstack.


bhufmann (approved):
LGTM. Thanks for the migration of the tests from the incubator repo.

PR #56, opened 2024-04-09
profiling.core: verify that the source flamechart is completed

arfio:
This patch corrects an issue where when opening a trace or experiment with the flamegraph view opened, it would schedule that call graph analysis before the call stack analysis was completed. This would result in an incorrect callgraph analysis and the flamegraph view would not work.

This issue was solved by checking that the call stack analysis is completed before calculating the callgraph.

[Fixed] Intermittent error with the flamegraph analysis


bhufmann (approved):
LGTM. Thanks for the contribution.

PR #59, opened 2024-04-11
tmf: Fix missing refresh due to stale sampling

### LINKED ISSUE #62: Missing time graph refresh
In some circumstances, time graph entries are missing their fine-detailed events and they are not refreshed when they should.

An example use case to reproduce this is to open two kernel traces that do not overlap in time, then select the first opened trace. Its Resources view time graph will be missing states, and only the coarse-detailed states are shown.

missing_states

PatrickTasse:
Fixes #62


bhufmann (commented):
It would be good to describe the bug in a issue tracker. Maybe have a little video animated gif for that. It's not clear to me what the issue is that it fixes.

PatrickTasse:
Fixes #62

bhufmann (approved):
Looks good to me. I fixes the bug reported. Thanks!

PR #60, opened 2024-04-11
ctf.core: Explicitly align to support 0-length sequences

### LINKED ISSUE #58: Parsing of CTF traces with 0-length sequence fields may fail due to alignment handling
See this LTTng issue, which includes a link to a reproducer: https://bugs.lttng.org/issues/1412

This happens for example with a trace collected on an aarch64 system.

christophebedard:
Resolves #58

For a test trace that covers this, see https://github.com/eclipse-tracecompass/tracecompass-test-traces/pull/3

This fix works for me with the above trace. It is based on the proposed babeltrace 1.5 fix (https://bugs.lttng.org/attachments/596) under the original LTTng issue (https://bugs.lttng.org/issues/1412).

I think that this fix is not needed for arrays, because the metadata parser simply rejects arrays with a length of less than 1 (i.e., throws an exception).

### DIFF
--- a/ctf/org.eclipse.tracecompass.ctf.core/src/org/eclipse/tracecompass/internal/ctf/core/event/types/SequenceDeclaration.java
+++ b/ctf/org.eclipse.tracecompass.ctf.core/src/org/eclipse/tracecompass/internal/ctf/core/event/types/SequenceDeclaration.java
@@ -121,6 +121,9 @@ public class SequenceDeclaration extends CompoundDeclaration {
             throw new CTFException("Sequence length too long " + length); //$NON-NLS-1$
         }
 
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Thanks!

christophebedard:
I believe that `org.eclipse.tracecompass.ctf.core.tests.trace.TraceReadAllTracesTest` should pick up the trace from https://github.com/eclipse-tracecompass/tracecompass-test-traces/pull/3 and try to read it (once a new release of `tracecompass-test-traces` is created), so there should be no need to add any new test, but please correct me if I'm wrong!

bhufmann (approved):
Thanks for the fix.

PR #61, opened 2024-04-11
xml: move xml callstack analysis to mainline

arfio:
This patch adds the callstack xml implementation. Because the incubator callstack implementation extends the same name, there is a conflict and sometimes the xml incubator callstack is chosen over the mainline one, but this should disappear if we remove the incubator callstack from the release.

A few modifications were made to allow for null values to be written with the rhino engine, because it would create error for undefined state values in the xml analyses.


tmf/org.eclipse.tracecompass.tmf.analysis.xml.core/META-INF/MANIFEST.MF
```diff
@@ -14,6 +14,7 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.tracecompass.datastore.core,
  org.eclipse.tracecompass.segmentstore.core,
  org.eclipse.tracecompass.analysis.timing.core,
```
bhufmann, line 16: The Bundle-Version above needs to be increased to 4.2.0.qualifier

tmf/org.eclipse.tracecompass.tmf.analysis.xml.core/src/org/eclipse/tracecompass/tmf/analysis/xml/core/module/TmfXmlStrings.java
```diff
@@ -215,6 +215,52 @@ public interface TmfXmlStrings {
 
     String PATTERN_HANDLER = "patternHandler";
 
+    /* XML callstack elements */
+    /**
+     * @since 4.1
```
bhufmann, line 220: all the versions for added APIs here and below should be 4.2

tmf/org.eclipse.tracecompass.tmf.analysis.xml.core/src/org/eclipse/tracecompass/tmf/analysis/xml/core/module/TmfAnalysisModuleHelperXml.java
```diff
@@ -60,6 +60,11 @@ public enum XmlAnalysisModuleType {
          * Analysis will be of type XmlPatternAnalysisModule
          */
         PATTERN,
+        /**
+         * Analysis will be XmlCallstackAnalysisModule
+         * @since 4.1
```
bhufmann, line 65: 4.2

bhufmann (changes requested):
There is also a test case that needs to be migrated. Please add that to the PR.

PR #63, opened 2024-04-17
releng: update alpine to have latest mvn

arfio:
Update the dockerfile because mvn is not up to date to build Trace Compass.

### DIFF
--- a/releng/dockerfiles/Dockerfile
+++ b/releng/dockerfiles/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 Polytechnique de Montréal
+# Copyright (c) 2023, 2024 Polytechnique de Montréal
 #
 # All rights reserved. This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0 which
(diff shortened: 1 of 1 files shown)


releng/dockerfiles/Dockerfile
```diff
@@ -17,6 +17,7 @@ WORKDIR /app/
 
 RUN mvn clean install -DskipTests -Dskip-rcp -Pfilter-grammar -Pctf-grammar
 
-FROM alpine:3.16.0
+FROM alpine:3.18.0
+
+COPY --from=packager /root/.m2/ /root/.m2/
```
bhufmann, line 22: Just curious, why is the whole .m2 directory copied now instead of the just the tracecompass sub-directory. Please let me know.

bhufmann (approved):
Looks good to me. Thanks!

releng/dockerfiles/Dockerfile
arfio: The whole directory is copied because some files outside of that are needed to build the incubator and the trace-server.

PR #64, opened 2024-04-25
TmfAnalysisElement: Execute analysis in separate thread

bhufmann:
This was done in UI thread and if module.schedule blocks due to waitForCompletion (e.g. of dependent analysis) the UI thread blocks until it's finished.


bhufmann:
Well, the schedule() call should not block. Calling stateSystem.waitForCompletion() should not be done in the analysis module. The analysis module should be fixed rather than workaround it. For that putting this a draft PR for now.

PR #66, opened 2024-04-28
Execution comparison : required changes on trace compass core for execution comparison analysis

farajidaneshgar:


tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/tmf/ui/viewers/xychart/TmfXYChartViewer.java
```diff
@@ -805,4 +805,13 @@ private void updateTimeFormat() {
             fDataProvider.setTimeFormat(fTimeFormat);
         }
     }
+    /**
+     * @since 9.1
+     * refresh fMouseSelectionProvider
+     */
+
+    public void refreshMouseSelectionProvider() {
```
arfio, line 813: I do not think this change is justified, the selectionRangeUpdated should update the mouse selection provider.

arfio (changes requested):
Some of those API changes need to be justified more clearly, otherwise the patch looks good and it may even be better if some tests were added !

tmf/org.eclipse.tracecompass.tmf.ui/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
-Bundle-Version: 9.0.3.qualifier
+Bundle-Version: 9.1.0.qualifier
```
arfio, line 5: This should be removed since you do not change the tmf plugin anymore.

analysis/org.eclipse.tracecompass.analysis.profiling.core/src/org/eclipse/tracecompass/internal/analysis/profiling/core/callgraph2/AggregatedCalledFunction.java
```diff
@@ -293,4 +293,24 @@ public AggregatedCalledFunctionStatistics getFunctionStatistics() {
     public String toString() {
         return "Aggregate Function: " + getObject() + ", Duration: " + getDuration() + ", Self Time: " + fSelfTime + " on " + getNbCalls() + " calls"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
     }
+    /**
```
arfio, line 296: There should a space in between those two functions. The file should be reformatted using eclipse format tool (ctrl+shift+f)

vladarama (commented):
Spoke with @farajidaneshgar offline and this change is not required anymore.
We can close this PR.

PR #69, opened 2024-04-30
tmf: Fix addMenuDetectListener() change of behavior in 4.32

PatrickTasse:
In TmfVirtualTable, we override Widget.addListener() to defer the listener to the Table child of the Composite.

In 4.32, Control.addMenuDetectListener() implementation changed from calling addListener() to calling the new protected method addTypedListener().

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1112

Since we can't override properly this new method (we can override it but the override implementation can't call the protected method on another Control), we must override addMenuDetectListener() instead.


bhufmann (approved):
Thanks for this! LGTM

PR #72, opened 2024-05-01
releng: Add staging/feature.xml

PatrickTasse:


bhufmann (approved):
Looks good to me. I tested it locally. Thanks!

PR #78, opened 2024-05-07
linux.core: Make getStatusIntervalsForThreads handle out of range queries

MatthewKhouzam:
This avoids many TimeRangeExceptions and undefined behaviour. It is visible when a kernel trace partially overlaps a userspace trace with a new callstack. The behavour is now defined rather than crashing out with an exception.


arfio (approved):
Looks good, thank you ! Typically, you would have kernel tracing before and after so I did not think about this issue when first implementing this.

MatthewKhouzam:
I tried making 2 kernel traces, and 2 ust traces in one experiment. That was not fun. :)

PR #85, opened 2024-05-17
tmf: Fix stream orElse(null) errors

PatrickTasse:
It seems that filter(Objects::nonNull) now changes the stream type to a <@NonNull E>, so that orElse(null) is no longer allowed. Change to check the optional presence before returning null.


MatthewKhouzam (approved):
I find this cleaner.

bhufmann (approved):
LGTM. Thanks!

Issue #87, opened 2024-05-22
Remove deprecated classes/APIs for Trace Compass 10.0


PatrickTasse:
This issue is to track the removal of deprecated classes/APIs in Trace Compass for the major 10.0 release. APIs can be removed that were deprecated before the 9.0 release for 2023-06.

Any API removal needs to reference this issue in the PR commit. 

If there are any issues with any removal of a specific API, please document it in this issue so that appropriate actions can be taken by Trace Compass project (e.g. bring the API back).

PR #91, opened 2024-05-23
analysis.graph: Cleanup non-API deprecation issues in dependency graph

PatrickTasse:
Fixes issue #87

[Removed] CriticalPathPalette


analysis/org.eclipse.tracecompass.analysis.graph.core/src/org/eclipse/tracecompass/internal/analysis/graph/core/criticalpath/CriticalPathModule.java
bhufmann, line 1: I think this class can be removed. Only the analysis ID constant is needed and needs to be moved to the abstract class. Also, the abstract class needs to extend TmfAbstractAnalysisModule and implement ICriticalPathProvider

bhufmann (changes requested):

bhufmann (approved):
Looks good to me. Thanks!

PR #92, opened 2024-05-23
releng: Add and fix create-new-and-noteworthy.py

PatrickTasse:
Fix to output markdown instead of mediawiki
Fix for removal of Change-Id from commit messages

Original copy taken from tracecompass.incubator


bhufmann (approved):
Looks good to me. Thanks!

PR #93, opened 2024-05-27
doc: improve development setup instructions

arfio:
- The link for incubator is not up to date.
- The instructions did not specify that the incubator target should be selected if building with the incubator.


bhufmann (approved):
Looks good to me. Thanks!

PR #102, opened 2024-06-12
releng: Update OOMPH setup file for GitHub

PatrickTasse:


bhufmann (approved):
Looks good. Do we need to update stable branch as well?

PR #104, opened 2024-06-14
Manage data provider factories

bhufmann:
tmf: Add API to retrieve data provider factories from dp manager
tmf: Add APIs to add and remove data provider factories

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
@@ -34,6 +34,7 @@ import org.eclipse.tracecompass.tmf.core.signal.TmfTraceClosedSignal;
 import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
 import org.eclipse.tracecompass.tmf.core.trace.TmfTraceManager;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimap;
 
@@ -248,5 +249,31 @@ public class DataProviderManager {
      */
     public <T extends ITmfTreeDataProvider<? extends ITmfTreeDataModel>> boolean removeDataProvider(ITmfTrace trace, T provider) {
         return fInstances.remove(trace, provider);
+    }
+
+    /**
+     * Get all registered data provider factories.
+     *
+     * @return a collection of existing data provider factories
+     * @since 9.4
+     */
+    public synchronized Collection<IDataProviderFactory> getFactories() {
+        return ImmutableList.copyOf(fDataProviderFactories.values());
+    }
+
+    /**
+     * Get a registered data provider factory for a given ID if it exists.
+     *
+     * @param id
+     *            The ID of the data provider factory or a data provider ID of
+     *            form factoryId:secondaryId
+     *
+     * @return the data provider factory for a given Id.
+     * @since 9.4
+     */
+    public synchronized @Nullable IDataProviderFactory getFactory(String id) {
+        String[] ids = id.split(DataProviderConstants.ID_SEPARATOR, 2);
+        String factoryId = ids.length > 1 ? ids[0] : id;
+        return fDataProviderFactories.get(factoryId);
     }
 }
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -34,7 +35,6 @@ import org.eclipse.tracecompass.tmf.core.signal.TmfTraceClosedSignal;
 import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
 import org.eclipse.tracecompass.tmf.core.trace.TmfTraceManager;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimap;
 
@@ -258,7 +258,7 @@ public class DataProviderManager {
      * @since 9.4
      */
     public synchronized Collection<IDataProviderFactory> getFactories() {
-        return ImmutableList.copyOf(fDataProviderFactories.values());
+        return List.copyOf(fDataProviderFactories.values());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -272,4 +272,53 @@ public synchronized Collection<IDataProviderFactory> getFactories() {
...
+     * dispose of it if needed.
+     *
+     * @param trace
+     *            The trace for which to remove the data provider
+     * @param id
+     *            The The ID of the data provider to remove
+     * @return {@code true} if any elements were removed
+     * @since 9.4
+     */
+    public boolean removeDataProvider(ITmfTrace trace, String id) {
```
PatrickTasse, line 306: Would it be reasonable to consider that when we replace or remove a data provider factory, then all instances that were created with that id would be automatically removed from here (for every trace)?

Also I wonder since we have no control over who has a reference to those data providers, how could any entity be sure that it would be ok to dispose of it? Probably just the owner of the data provider, who would be unaware that its factory was ever removed. It can continue to use the data provider until it no longer needs it?

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -249,4 +250,26 @@ public List<IDataProviderDescriptor> getAvailableProviders(@Nullable ITmfTrace t
     public <T extends ITmfTreeDataProvider<? extends ITmfTreeDataModel>> boolean removeDataProvider(ITmfTrace trace, T provider) {
         return fInstances.remove(trace, provider);
     }
+
+    /**
```
PatrickTasse, line 254: Javadoc description

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -249,4 +250,26 @@ public List<IDataProviderDescriptor> getAvailableProviders(@Nullable ITmfTrace t
...
+     * @return a collection of existing data provider factories
+     * @since 9.4
+     */
+    public synchronized Collection<IDataProviderFactory> getFactories() {
+        return ImmutableList.copyOf(fDataProviderFactories.values());
+    }
+
+    /**
```
PatrickTasse, line 262: Javadoc description

PatrickTasse (dismissed):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -272,4 +272,53 @@ public synchronized Collection<IDataProviderFactory> getFactories() {
...
+     * If a data provider factory is associated with the ID already exists
+     * it will replace and return the existing data provider factory.
+     *
+     * @param id
+     *            The data provider factory ID
+     * @param factory
+     *            The data provider factory implementation
+     * @return the previous data provider factory associated with the key or null
+     *
+     * @since 9.4
```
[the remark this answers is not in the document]
bhufmann: Done

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -272,4 +272,53 @@ public synchronized Collection<IDataProviderFactory> getFactories() {
         String factoryId = ids.length > 1 ? ids[0] : id;
         return fDataProviderFactories.get(factoryId);
     }
+
+    /**
+     * Add a new data provider factory.
+     *
+     * If a data provider factory is associated with the ID already exists
```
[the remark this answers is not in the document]
bhufmann: Done

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
bhufmann: > Would it be reasonable to consider that when we replace or remove a data provider factory, then all instances that were created with that id would be automatically removed from here (for every trace)?
> 
I implemented it the same way as the other removeDataProvider() method (line 250) which is used with ScriptedDataProviderManager. The new method is used for similar use case: Some entity is creating data provider which is registered to the DataProviderManager and the same entity is responsible to remove it. 

I get your point. Any suggestion how to implement that differently for this use case?

BTW, maybe we should return the original data provider if it exists else null, to be consistent to the other data provider method. WDYT?

> Also I wonder since we have no control over who has a reference to those data providers, how could any entity be sure that it would be ok to dispose of it? Probably just the owner of the data provider, who would be unaware that its factory was ever removed. It can continue to use the data provider until it no longer needs it?

No, we don't have any control over who has a reference on the data provider. As stated a few lines above, the use case is that an entity is create a data provider and is responsible to remove and dispose it. Please let me know if you have any suggestion on how to handle that use case.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
bhufmann: Done

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
bhufmann: Done

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
bhufmann: As per our chat, I'll remvoe the removeDataProvider(trace, id) since it's not needed, and I'll update add/removeDataProviderFactory() to remove data providers instances for all traces if there are conflicts.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -249,4 +250,30 @@ public List<IDataProviderDescriptor> getAvailableProviders(@Nullable ITmfTrace t
...
     }
+
+    /**
+     * Get all registered data provider factories.
+     *
+     * @return a collection of existing data provider factories
+     * @since 9.4
+     */
+    public synchronized Collection<IDataProviderFactory> getFactories() {
+        return ImmutableList.copyOf(fDataProviderFactories.values());
```
MatthewKhouzam, line 262: for future, List.of()?

MatthewKhouzam (approved):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
bhufmann: List.copyOf()

MatthewKhouzam (approved):
lgtm!

PR #75, opened 2024-05-06
analysis: Add "Timestamp ns" column in SegmentStoreTableDataProvder

bhufmann:
This will allow time synchronization when selecting a row in the Theia/VsCode extension to the trace server.

Fixes #74

[Added] "Timestamp ns" column in SegmentStoreTableDataProvder

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #111, opened 2024-06-17
analysis: Add "Timestamp ns" column in SegmentStoreTableDataProvder

### LINKED ISSUE #74: Add "Timestamp ns" column to the SegmentStoreTableDataProvider
This would enable time sychronization in the theia-trace-extension/vscode-trace-extension front-ends from the table views (similar to the Events Table). 

Fixes https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/1074

bhufmann:
This will allow time synchronization when selecting a row in the Theia/VsCode extension to the trace server.

It uses the same mechanism than the TmfEventsTableTableProvider using a ns timestamp column

Fixes #74

[Added] "Timestamp ns" column in SegmentStoreTableDataProvider


bhufmann:
Duplicated PR. See #75 instead.

PR #115, opened 2024-06-19
tmf: Allow TmfConfiguration to be de-/serialized with GSON

bhufmann:
Allow TmfConfiguration to be de-/serialized with GSON. 
Generate unique UUID as ID from name and parameters
Change parameter map to String->String for easy serialization.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfiguration.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfiguration.java
@@ -43,5 +43,5 @@ public interface ITmfConfiguration {
      * @return optional informational parameters to return. Can be used to show
      *         more details to users of the configuration instance.
      */
-    Map<String, Object> getParameters();
+    Map<String, String> getParameters();
 }
--- a/tmf/org.eclipse.tracecompass.tmf.core.tests/META-INF/MANIFEST.MF
+++ b/tmf/org.eclipse.tracecompass.tmf.core.tests/META-INF/MANIFEST.MF
@@ -63,5 +63,8 @@ Export-Package: org.eclipse.tracecompass.tmf.core.tests,
  org.eclipse.tracecompass.tmf.tests.stubs.trace.xml
 Import-Package: com.google.common.base,
  com.google.common.collect,
- com.google.common.primitives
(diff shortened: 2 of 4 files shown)


MatthewKhouzam (commented):
LGTM!

MatthewKhouzam (approved):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfiguration.java
```diff
@@ -43,5 +43,5 @@ public interface ITmfConfiguration {
      * @return optional informational parameters to return. Can be used to show
      *         more details to users of the configuration instance.
      */
-    Map<String, Object> getParameters();
+    Map<String, String> getParameters();
```
bhufmann, line 46: Some design discussion is needed for this because it changes the parameter to a map<String, String> instead of map<String, Object>. This allows to serialize the object from the base class and there is no need for each extension to implement their own serialization. However, this removes the possibility to have more complex values, e.g. arrays or built-in types like requested_items. Besides it's not clear when using requested_items as parameters how to serialize, because these values are not REST-ful and changes between executions. Hence, other data needs to be serialized instead of the current requested_items values (maybe use tree path).

MatthewKhouzam: I kinda think that gson would serialize like
"item: { a: [1,2,3,4,5,6]} " No?

bhufmann: If we change value to be a string, then FE can't use an array as value. The FE will have to provide a String with requested items encoded somehow. That's different than to other endpoints where we use requested_items. requested_items was just one example. Other TSP specified query-parameters have similar issues.

The other issue is that that class cannot blindly serialize the parameters as they come from the FE because some of them are not REST-Ful, example requested_items. Not clear how to handle that. For that we would need a custom serializer.

PatrickTasse (approved):

bhufmann:
Converted to draft because it's not clear if we will go ahead with a generic serialization feature or not, as well as because open questions.

bhufmann:
Closing it due the previously mentioned open issues. Will re-open if necessary.

PR #117, opened 2024-06-26
releng: Downgrade org.mozilla.rhino to org.mozilla.javascript

### LINKED ISSUE #116: EASE Scripting feature fails at initialization
The incubator EASE Scripting feature throws an exception at initialization (e.g. switching to Scripting perspective).

EASE (org.eclipse.ease) has a dependency on org.mozilla.javascript plug-in. Trace Compass is using the latest version of this feature which has been renamed to org.mozilla.rhino, for use with the XML Analysis feature.

When both plug-ins are installed it creates a conflict that causes initialization failure of the RhinoScriptEngine.

Trace Compass needs to downgrade to use the older org.mozilla.javascript plug-in to ensure compatibility with the incubator EASE Scripting feature. This older version can be found in previous Orbit release repositories.

To use the latest version of org.mozilla.rhino plug-in, it will require an update from the EASE project first.

PatrickTasse:
Use older version of org.mozilla.javascript for compatibility with incubator EASE Scripting feature.

Fixes #116

### DIFF
--- a/rcp/org.eclipse.tracecompass.rcp/feature.xml
+++ b/rcp/org.eclipse.tracecompass.rcp/feature.xml
@@ -230,7 +230,7 @@
          version="0.0.0"/>
 
    <plugin
-         id="org.mozilla.rhino"
+         id="org.mozilla.javascript"
          version="0.0.0"/>
 
    <plugin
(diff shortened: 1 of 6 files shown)


bhufmann (approved):
Looks good to me. Thanks!

PR #123, opened 2024-07-10
Event aspect data type

kavehshahedi:
Added data type to the event and segment aspects (Segment already had getType, but the aspects didn't). The aim was to synchronize the trace-server functionalities w.r.t. tables in order to indicate the type of columns.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfEventTableDataProvider.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfEventTableDataProvider.java
@@ -152,7 +152,7 @@ public class TmfEventTableDataProvider extends AbstractTmfTableDataProvider impl
         for (ITmfEventAspect<?> aspect : aspects.values()) {
             synchronized (fAspectToIdMap) {
                 long id = fAspectToIdMap.computeIfAbsent(aspect, a -> createColumnId());
-                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault()));
+                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault(), aspect.getDataType()));
                 hasTs |= (aspect == TmfBaseAspects.getTimestampAspect());
             }
         }
@@ -160,7 +160,7 @@ public class TmfEventTableDataProvider extends AbstractTmfTableDataProvider impl
             synchronized (fAspectToIdMap) {
                 ITmfEventAspect<Long> aspect = TmfBaseAspects.getTimestampNsAspect();
                 long id = fAspectToIdMap.computeIfAbsent(aspect, a -> createColumnId());
-                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault()));
+                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault(), aspect.getDataType()));
             }
         }
         return new TmfModelResponse<>(new TmfTreeModel<>(Collections.emptyList(), model), ITmfResponse.Status.COMPLETED, CommonStatusMessage.COMPLETED);
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfEventTableDataProvider.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfEventTableDataProvider.java
@@ -152,7 +152,7 @@ public class TmfEventTableDataProvider extends AbstractTmfTableDataProvider impl
         for (ITmfEventAspect<?> aspect : aspects.values()) {
             synchronized (fAspectToIdMap) {
                 long id = fAspectToIdMap.computeIfAbsent(aspect, a -> createColumnId());
-                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault(), aspect.getDataType()));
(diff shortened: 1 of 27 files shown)


MatthewKhouzam (commented):
Looks like a great approach to me!

MatthewKhouzam (approved):
LGTM

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfEventTableDataProvider.java
```diff
@@ -152,15 +152,15 @@ public TmfModelResponse<TmfTreeModel<TmfEventTableColumnDataModel>> fetchTree(Ma
         for (ITmfEventAspect<?> aspect : aspects.values()) {
             synchronized (fAspectToIdMap) {
                 long id = fAspectToIdMap.computeIfAbsent(aspect, a -> createColumnId());
-                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault()));
+                model.add(new TmfEventTableColumnDataModel(id, -1, Collections.singletonList(aspect.getName()), aspect.getHelpText(), aspect.isHiddenByDefault(), aspect.getDataType().name()));
```
bhufmann, line 155: This comment is for line 149 above where a `MultiAspect` can be created. A `MultiAspect` is created for aspects that are coming from the different traces and provide the same information, for example cpu, pid or tid. This avoids to have duplicated and redundant columns. The way how a `MuiltAspect` is created is that aspects have to have a common parent class. For example `LinuxTidAspect` is the base class for `KernelTidAspect` and  `ContextTidAspect`.

So, if one has an experiment with a LTTng UST and Kernel trace a `MultiAspect` is created and  currently it doesn't override the getDataType() method, and hence the data type that the `TmfEventTableDataProvider` provides is the default `STRING`.

Please update the `MultiAspect` to implement the `getDataType()` method and return the correct data type from the stored `fAspects` class variable (use the first one in the list). 
```java
@Override
    public DataType getDataType() {
        for (ITmfEventAspect<?> aspect : fAspects) {
            return aspect.getDataType();
        }
        return ITmfEventAspect.super.getDataType();
    }
```

Please also add a check in the `MultiAspect.create(aspects, aspectClass)` method if the list of aspects don't have all the same data type (which they should) similar to the warning if the names of aspects are not the same. See here what I mean https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/c8509f55bb32a5a7ad3ee7eb4ffe7d7828ac7316/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/event/aspect/MultiAspect.java#L80).

bhufmann (changes requested):
Thanks for this contribution. I have a few comments.

I noticed that you have multiple commits. That's ok, however, each commit should be a logical increment of the implemented feature, which has a successful build. The first 2 commits is for events aspects, the third one for event and segment store aspects and the last is to fix a broken test. One suggestion is to have event aspect changes in one commit and the segment store changes in another.

I also noticed that the commit messages are not formatted properly and missing signed-of line. Please also update the commit messages for that (see here for some guidelines: https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/master/CONTRIBUTING.md#pull-request-guidelines)

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/event/aspect/ITmfEventAspect.java
```diff
@@ -107,4 +108,16 @@ public interface ITmfEventAspect<T> {
...
+
+
+    /**
+     * This method will return the data type of the aspect.
+     * The data type is considered as String by default
+     *
+     * @return the data type of the aspect
+     * @since 9.3
+     */
+    default DataType getDataType() {
```
bhufmann, line 120: This method should be added also to `ISegmentAspect` interface and related implementations of that interface (See also SegmentStoreTableDataProvider that use `ISegmentAspect`)

bhufmann:
@kavehshahedi what's the status on this PR?

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/event/aspect/ITmfEventAspect.java
kavehshahedi: I investigated this suggestion, but it doesn't sound proper;

1. The `DataType getDataType()` in the `ITmfEventAspect` class returns the data type based on the `DataType` enum.
2. The `SegmentType getType()` in the `ISegmentAspect` class returns the segment type based on the `SegmentType` enum.
3. Your recommendation is to add a `DataType`-based method in the `ISegmentAspect` class to return the data type of the segment. However, the type that a segment can have is either `CATEGORICAL` or `CONTINUOUS`, which are not present in the `DataType` enum. 
4. As we need to have a consistent enum type for both classes to use in the `TmfTreeDataModel` class, your suggestion may not be suitable.

Hence, I was thinking of having a common enum interface for `DataType` and `SegmentType`, so we can handle it easier.

bhufmann: I think CATEGORICAL/CONTINUOUS and the values of DataType are complementary and not just one each. A column can have DataType NUMBER and can be CONTINUOUS or a column could have a DataType NUMBER but it's CATEGORICAL. So, you can't just have one enum which includes values of DataType and SegmentType. Or do I miss understand something?

@MatthewKhouzam any comment from your side?

kavehshahedi: Oh, I didn't know that a segment can have both types (data type + segment type) at the same time. In this case, that wouldn't be a problem, and your solution works fine. I guess @MatthewKhouzam can also verify it.

MatthewKhouzam (changes requested):
I agree with Bernd's take.

MatthewKhouzam (approved):
Type and DataType are different. I'm good with this!

@kavehshahedi  you need to update the commit messages with 
[added] an [changed] 

```If the patch is significant enough to have a mention in the New & Noteworthy, annotate the patch with [Fixed], [Added], [Changed], [Deprecated], or [Security] with a short description of what the change does.```

MatthewKhouzam (approved):

bhufmann (approved):
Thanks for the update. It looks good to me.

PR #125, opened 2024-07-16
Implement interface to delete persistant data for segment store analysis (on-disk)

bhufmann:
This PR contains 2 commits:

- Implement AbstractSegmentStoreAnalysisModule.clearPersistentData()
  - This allows to delete on-disk segment store data file from the analysis module. If segment store is open, it will close the segment store file first before deleting the file. Otherwise delete the file directly.
  - Calling methods are responsible to re-trigger the analysis execution
  - This allows deleting on-disk segment store state while a trace is open
- timing.tests: Move SegmentStoreAnalysisTest to stubs directory


MatthewKhouzam (approved):
Looks good to me.

MatthewKhouzam (approved):

PR #126, opened 2024-07-16
Trace event update

MatthewKhouzam:
Broken, but want to discuss


bhufmann:
The version of the trace-event-logger is 0.0.4. Should it be version 1.0 first, before using it in the Trace Compass project?

MatthewKhouzam:
I'm still in the testing phase. Thanks!

PR #134, opened 2024-08-16
tmf: Support list of values in matches statements in timegraph filter

### LINKED ISSUE #81: Time Graph filter doesn't support list of values in matches statements
The time graph filter doesn't support list of values in matches statements. For example, using the Resources view with a LTTng kernel trace, the filter string `CPU matches [0,1]` doesn't match anything and by default everything is shown and non-matching CPU rows are not dimmed. On the contrary, value ranges, like `CPU matches [0-1]` will highlight matching CPU rows and dim the rest. 

rcv-cpu-filtering-bug

bhufmann:
Add comma to list of supported characters in FilterParser ANTLR grammar.

Fixes #81

[Fixed] Support list of values in matches statements in timegraph filter


MatthewKhouzam (commented):
Do we need this still?

bhufmann:
> Do we need this still?

Maybe not for the list of values. Comma is not really needed. However, for qualifiers, e.g. `{2,4}` it would be useful?

MatthewKhouzam:
My question, is it deviating from regexes, and thus would it be less portable in the long run.
Let's say we move away from our grammar, would we need to keep the logic elsewhere?

bhufmann:
> My question, is it deviating from regexes, and thus would it be less portable in the long run. Let's say we move away from our grammar, would we need to keep the logic elsewhere?

`,` as valid text character was missing before. Because of that, it was not possible to match a comma or use it in regex quantifiers. So, I still think it should be added. Maybe I need to change the commit and PR description to make it clear.

MatthewKhouzam:
Yes, my question is that [0,2] is not really a regex, I think. It will only match the characters '0', ',' and '2'. Do we want that?

bhufmann:
> Yes, my question is that [0,2] is not really a regex, I think. It will only match the characters '0', ',' and '2'. Do we want that?

`[0,2]` is a regex. It looks for `0` or `,` or `2`.

MatthewKhouzam (approved):
I believe this is the right way forward too!

PR #140, opened 2024-08-30
Add default values to SegmentStoreStatisticsAspects

siwei-zhang-work:
The statistics previously used default values like Double.MAX_VALUE and Double.MIN_VALUE for fields such as min and max, which were displayed when no elements were present in the statistics. This commit introduces a default value of null for situations where there are no elements in the statistics.

[Added] Default value for SegmentStoreStatisticsAspects


bhufmann (approved):
LGTM. Thanks!

MatthewKhouzam (approved):
Looks good to me.

[Added] instead of [Add]

PR #141, opened 2024-08-30
tmf: Add support for json-schema file to TmfConfigurationSourceType

bhufmann:
For schema specification see https://json-schema.org/

[Added] support for json-schema file to TmfConfigurationSourceType

See https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/103

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
@@ -11,7 +11,10 @@
 
 package org.eclipse.tracecompass.tmf.core.config;
 
+import java.io.File;
 import java.util.List;
+
+import org.eclipse.jdt.annotation.Nullable;
 
 /**
  * Interface to implement that describes a configuration source.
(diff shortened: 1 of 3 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
```diff
@@ -49,4 +52,16 @@ public interface ITmfConfigurationSourceType {
      * @return A list of query parameter descriptors to be passed
      */
     List<ITmfConfigParamDescriptor> getConfigParamDescriptors();
+    /**
```
PatrickTasse, line 55: Missing blank line before

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
```diff
@@ -49,4 +52,16 @@ public interface ITmfConfigurationSourceType {
...
+     * A string containing a json-schema describing the query parameters
```
PatrickTasse, line 56: Gets a file containing...

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
```diff
@@ -49,4 +52,16 @@ public interface ITmfConfigurationSourceType {
...
+     * to be passed when creating a configuration instance of this type.
+     *
+     * Note: Use either {@link #getConfigParamDescriptors()} or {@link #getSchemaFile()}
```
PatrickTasse, line 59: For the user, they don't know which is to be used. I would suggest:
Note: If the schema file is null, use {@link #getConfigParamDescriptors()} instead.

Should we also add a note in that other method?
Note: Use this method if {@link #getSchemaFile()} returns null.

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfConfigurationSourceType.java
bhufmann: Done

PatrickTasse (approved):

PR #144, opened 2024-09-06
Json schema json if

bhufmann:
This PR includes changes of PR #141.

An additional commit is added:
- tmf: Add API constant for retrieving json string from TmfConfiguration

```
public static final String JSON_STRING_KEY = "--json-string-key--";
```
This is a compromise to not having to depend on a json parsing library in the ITmfConfiguration interface.


bhufmann:
Not needed anymore, because only Map<String, Object> was decided to be used (see here: https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/pull/88#discussion_r1791829115)

PR #154, opened 2024-09-23
tmf: add ITmfDataProviderConfigurator interface for custom data provider

bhufmann:
Add interface to implement to prepare a derived data provider with input parameter defined by ITmfConfigurationSourceType.

For that add interfaces ITmfDataProviderConfigurator for capabilities to create data providers that a data provider factory could implement.

Add getAdapter() interface to IDataProviderFactory to allow retrieving ITmfDataProviderConfigurator.

Example:
ITmfDataProviderConfigurator configurator =
factory.getAdapter(ITmfDataProviderConfigurator.class);

[Added] ITmfDataProviderConfigurator interface for custom data provider
[Added] getAdapter() interface to IDataProviderFactory

For example, to get a ITmfDataProviderSource implementation do:
```java
   ITmfDataProviderSource source = factory.getAdapter(IDataProviderSource.class);
```

Note that this PR includes changes of other PR (#141)

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfDataProviderConfigurator.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfDataProviderConfigurator.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Ericsson
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License 2.0 which
+ * accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.tracecompass.tmf.core.config;
+
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.tracecompass.tmf.core.dataprovider.DataProviderManager;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderDescriptor;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfConfigurationException;
+import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
+
+/**
+ * Interface for creating data providers. Implementers only return a data provider descriptor.
+ * Instantiating the actual data provider(s) will be done using {@link DataProviderManager}.
+ * @since 9.5
+ */
+public interface ITmfDataProviderConfigurator {
+
+    /**
+     * @return a list of {@link ITmfConfigurationSourceType}
+     */
+    List<ITmfConfigurationSourceType> getConfigurationSourceTypes();
+
+    /**
+     * Prepares a data provider based on input parameters and returns its
+     * corresponding data provider descriptor.
+     *
+     * The data provider descriptor shall return the parent data provider ID
+     * through {@link IDataProviderDescriptor#getParentId()}, as well as the
+     * creation configuration through
+     * {@link IDataProviderDescriptor#getConfiguration()}.
+     *
+     * @param typeId
+     *            The Configuration type ID specified in corresponding
+     *            {@link ITmfConfigurationSourceType}
+     * @param trace
+     *            The trace (or experiment) instance
+     * @param parameters
+     *            The configuration parameters.
+     * @return a data provider descriptor corresponding to the derived data
+     *         provider.
+     *
+     * @throws TmfConfigurationException
+     *             if an error occurs
+     */
+    IDataProviderDescriptor createDataProviderDescriptors(String typeId, ITmfTrace trace, Map<String, Object> parameters) throws TmfConfigurationException;
+
+    /**
+     * Remove a data provider provider that was created by
+     * {@link #createDataProviderDescriptors(String, ITmfTrace, Map)}
+     *
+     * @param trace
+     *            The trace (or experiment) instance
+     * @param descriptor
+     *            The data provider descriptor
+     * @throws TmfConfigurationException
+     *             if an error occurs
+     */
+    void removeDataProviderDescriptor(ITmfTrace trace, IDataProviderDescriptor descriptor) throws TmfConfigurationException;
+}
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderFactory.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderFactory.java
@@ -14,8 +14,10 @@ package org.eclipse.tracecompass.tmf.core.dataprovider;
 import java.util.Collection;
 import java.util.Collections;
 
+import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.tracecompass.tmf.core.config.ITmfDataProviderConfigurator;
 import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataModel;
 import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataProvider;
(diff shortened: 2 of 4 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfDataProviderSource.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param trace
+     *            The trace (or experiment) instance
+     * @param jsonParameters
+     *            The configuration parameters as valid json string
+     * @return List of data provider descriptor
+     * @throws TmfConfigurationException
+     *             if an error occurs
+     */
+//    List<IDataProviderDescriptor> createDataProviderDescriptors(String typeId, ITmfTrace trace, Map<String, Object> parameters) throws TmfConfigurationException;
+    List<IDataProviderDescriptor> createDataProviderDescriptors(String typeId, ITmfTrace trace, String jsonParameters) throws TmfConfigurationException;
```
bhufmann, line 41: The jsonParameter contains the json as string that was specified in configuration source type. I requires a name field and an optional description field. The name will be shown to the user as part of the data provider descriptor it creates.

Notes:

That this interface means that the jsonParameter are assigned to all traces in the experiment. It's not required at the moment to have parameters per trace.

The data providers and descriptors need an ITmfTrace instance in TC.core, so we need to pass the instance. Ideally for configurations, it shouldn't be necessary to have a trace object (that also executes analysis). 

To avoid that there would need to be a separate model representing traces/experiments/analysis like it is done with classic Trace Compass UI model. Refactoring for that would be big undtertaking.

abhinava-ericsson:
I found the "*Source" classes to be some kind of a catch-all which doesn't convey much. I would suggest the following:

1. Renaming (I)TmfDataProviderSource to (I)TmfDataProviderCustomizer
2. Renaming (I)TmfConfigurationSourceType to (I)TmfConfigurationType
3. Renaming (I)TmfConfigurationSource to (I)TmfConfigurationFactory
4. Renaming TmfConfigurationSourceManager to TmfConfigurationManager

In case 2-4 have already been exposed in Trace Compass API, 1 alone would still be helpful.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderFactory.java
```diff
@@ -77,4 +78,16 @@ public interface IDataProviderFactory {
...
+     *  Disposes the data provider factory
+     *  @since 9.5
+     */
+    default void dispose() {
+        // do nothing
+    }
+
+    @Override
+    default <T> T getAdapter(Class<T> adapter) {
+        return null;
```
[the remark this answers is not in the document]
bhufmann: Thanks for the suggestion!

abhinava-ericsson:
> The `TmfConfigurationSourceManager` manages instances of `TmfConfigurationSource` defined as extensions to the corresponding extension point. Each `TmfConfigurationSource` has a type `TmfConfigurationSourceType` that describes the source. Each source is then managing instances of `ITmfConfiguration`. Hence the term `TmfConfigurationManager` as replacement to `TmfConfigurationTypeManager` would not fit, because each source is managing its configurationsm and hence are the manager of configurations. I chose the different name `TmfConfigurationTypeManager` to distinguish this different difference

Sorry, for 4, I meant to say:
- Renaming TmfConfigurationSourceManager to TmfConfigurationTypeManager.

But, if we are sticking with Source for 2 and 3, I guess it doesn't matter.

bhufmann:
> > The `TmfConfigurationSourceManager` manages instances of `TmfConfigurationSource` defined as extensions to the corresponding extension point. Each `TmfConfigurationSource` has a type `TmfConfigurationSourceType` that describes the source. Each source is then managing instances of `ITmfConfiguration`. Hence the term `TmfConfigurationManager` as replacement to `TmfConfigurationTypeManager` would not fit, because each source is managing its configurationsm and hence are the manager of configurations. I chose the different name `TmfConfigurationTypeManager` to distinguish this different difference
> 
> Sorry, for 4, I meant to say:
> 
>     * Renaming TmfConfigurationSourceManager to TmfConfigurationTypeManager.
> 
> 
> But, if we are sticking with Source for 2 and 3, I guess it doesn't matter.

TmfConfigurationTypeManager would make sense as a name. "Source" is not really needed in the class name. 

It's already API, so we would adhere to the Trace Compass API policy for breaking changes. However, we could take the stand that these APIs are still provisional (even if not marked as provisional) and it's not used outside Trace Compass Trace Server. In both cases changing it it will require changes  in multiple places including TSP. I'll discuss with it the team first and see what we are going to do next.

bhufmann:
> > I found the "*Source" classes to be some kind of a catch-all which doesn't convey much. I would suggest the following:
> > 
> > 1. Renaming (I)TmfDataProviderSource to (I)TmfDataProviderCustomizer
> > 2. Renaming (I)TmfConfigurationSourceType to (I)TmfConfigurationType
> > 3. Renaming (I)TmfConfigurationSource to (I)TmfConfigurationFactory
> > 4. Renaming TmfConfigurationSourceManager to TmfConfigurationManager
> > 
> > In case 2-4 have already been exposed in Trace Compass API, 1 alone would still be helpful.
> 
I thought about a bit more. For 1) what about `(I)TmfDataProviderConfigurator`. Customizer for would mean that an existing data provider is customized. @abhinava-ericsson @PatrickTasse  WDYT?

Renaming 2-4 would mean a fullstack update (BE, FEs, TSP, Trace Compass), incl. following the API deprecation policy of Trace Compass and this would be quite effort for little gain.

abhinava-ericsson:
> > > I found the "*Source" classes to be some kind of a catch-all which doesn't convey much. I would suggest the following:
> > > 
> > > 1. Renaming (I)TmfDataProviderSource to (I)TmfDataProviderCustomizer
> > > 2. Renaming (I)TmfConfigurationSourceType to (I)TmfConfigurationType
> > > 3. Renaming (I)TmfConfigurationSource to (I)TmfConfigurationFactory
> > > 4. Renaming TmfConfigurationSourceManager to TmfConfigurationManager
> > > 
> > > In case 2-4 have already been exposed in Trace Compass API, 1 alone would still be helpful.
> 
> I thought about a bit more. For 1) what about `(I)TmfDataProviderConfigurator`. Customizer for would mean that an existing data provider is customized. @abhinava-ericsson @PatrickTasse WDYT?
> 
> Renaming 2-4 would mean a fullstack update (BE, FEs, TSP, Trace Compass), incl. following the API deprecation policy of Trace Compass and this would be quite effort for little gain.

`(I)TmfDataProviderConfigurator` looks good to me.
Leaving 2-4 as is Ok; just the classes with "Source" in their names should be documented to help a developer parse what "Source" really means in the context of each class.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderFactory.java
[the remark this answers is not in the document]
bhufmann: Done

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfDataProviderConfigurator.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public interface ITmfDataProviderConfigurator {
+
+    /**
+     * @return a list of {@link ITmfConfigurationSourceType}
+     */
+    List<ITmfConfigurationSourceType> getConfigurationSourceTypes();
+
+    /**
+     * Prepares a data provider based on input parameters and returns it's
```
PatrickTasse, line 35: its

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/ITmfDataProviderConfigurator.java
bhufmann: Done

PatrickTasse (approved):

PR #155, opened 2024-09-23
Expose the clock properties in CTFTrace

siwei-zhang-work:
There is the need to get clock scale of the trace.
This commit added the clock scale to the environment variables of the trace.

[Added] clock scale in getEnvironment() in CTFTrace
[Changed] moved clock offset into getEnvironment() in CTFTrace
[Deprecated] CtfTmfTrace.CLOCK_OFFSET, use CTFTrace.CLOCK_OFFSET instead.

Signed-off-by: Siwei Zhang <siwei.zhang@ericsson.com>

### DIFF
--- a/ctf/org.eclipse.tracecompass.ctf.core.tests/src/org/eclipse/tracecompass/ctf/core/tests/trace/MetadataTest.java
+++ b/ctf/org.eclipse.tracecompass.ctf.core.tests/src/org/eclipse/tracecompass/ctf/core/tests/trace/MetadataTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2014 Ericsson
+ * Copyright (c) 2013, 2024 Ericsson
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


MatthewKhouzam:
Hi Siwei, any thoughts on complementing getEnvironment with the frequency? It would require less code and no API changes.

siwei-zhang-work:
> Hi Siwei, any thoughts on complementing getEnvironment with the frequency? It would require less code and no API changes.

But the getEnvironment() refers to a specific `env` block in the metadata, and complementing getEnvironment() still requires exposing frequency from the CTFClock. I changed to adding the frequency in the getProperties(), and the frequency is exposed the same way as time offset. Does this seem cleaner? Otherwise I'll add the frequency in the getEnvironment().

MatthewKhouzam:
getEnvironment currently refers to `env` but it means all data that is not in events. I would recommend using it for all static information, also I will add get environment to TmfTrace. Trace Event should have this for metadata too. Maybe we can get a second opinion, but I would not add a new api for this instead of enhancing the output.

siwei-zhang-work:
> getEnvironment currently refers to `env` but it means all data that is not in events. I would recommend using it for all static information, also I will add get environment to TmfTrace. Trace Event should have this for metadata too. Maybe we can get a second opinion, but I would not add a new api for this instead of enhancing the output.

Sounds good, done.

MatthewKhouzam (approved):

MatthewKhouzam (changes requested):
Please make sure you're not breaking the API. @deprecate instead.

MatthewKhouzam (approved):

bhufmann (approved):
LGTM. Thanks!

PR #157, opened 2024-09-23
statesystem: Do not reduce time range condition for root nodes

### LINKED ISSUE #156: HistoryTreeBackendIterator skips nodes during construction
In HistoryTreeBackendIterator, the time range condition for queuing children nodes is reduced according to the current node time range.

During construction, it can happen that the root node has a time range that is only the start timestamp. In that case the iterator will only consider child nodes that intersect with the start time and all other child nodes will be discarded. The iterator can then miss many intersecting intervals.

The root node time range can be increased if intervals are added to it (this does not always happen) or at the end of construction when the node is closed.

A solution seems to be to avoid reducing the time range condition when the current node is the root node.

PatrickTasse:
In HistoryBackEndIterator, do not reduce the time range condition for queuing children nodes when the current node is the root node, as its time range is not finalized during construction.

Fixes #156

[Fixed] Do not reduce time range condition for root nodes


arfio (approved):
If the node checked is the parent node, then it should not restrict the query as it will always match the last event timestamp. In the case of CTF, I noticed that the end time of the trace does not match the last event timestamp, that is due to the fact that the end timestamp of the trace is set according to packet end times which are usually "packet flushing timestamp".

In any case, this change looks good to me ! Thank you.

PR #159, opened 2024-09-23
statesystem: Implement new query2D API in ThreadedHistoryTreeBackend

### LINKED ISSUE #158: ThreadedHistoryTreeBackend missing intervals in query2D
In https://git.eclipse.org/r/134665 a 'reverse' parameter was added to the signature of the IStateHistoryBackend.query2D() interface.

The state system was changed to call this new method. But ThreadedHistoryTreeBackend was not updated to implement this method, and therefore the implementation in base class HistoryTreeBackend is used. The base class is not aware of and misses intervals that are currently held in the interval queue (BufferedBlockingQueue) waiting to be written to file.

PatrickTasse:
The new API for query2D with reverse parameter is implemented in ThreadedHistoryTreeBackend. This avoids the base class implementation from being used, which is missing queued intervals.

Fixes #158

[Fixed] Implement new query2D API in ThreadedHistoryTreeBackend\


arfio (approved):
LGTM !

PR #162, opened 2024-09-30
profiling: Change DP name of Flame Charts created by different analyses

bhufmann:
Prepend the analysis name before the view type, e.g. "Function Graph - Flame Chart". This to be consistent with other data providers that have common analysis (e.g. Function Graph - Latency Statistics)

This helps also with the default grouping by name of data providers in the trace server front-end.

[Changed] DP name of Flame Chart created by different analyses


bhufmann:
Tested with Ftrace for Function Graph Flame chart. Note that for other analysis the Flame chart and latency analysis can be further apart due sorting by name.

Before:

image

After:
image

MatthewKhouzam (dismissed):

PatrickTasse (approved):

PR #161, opened 2024-09-30
tmf: Add API to add and remove analysis modules to/from ITmfTrace

bhufmann:
Many analysis utilities require analysis modules to be present in the ITmfTrace.getAnalysisModules() collection. This one is filled using the TMF analysis framework (see IAnalysisModuleSource implementation). The analysis framework will populate available IAnalysisModuesHelpers at startup (plug-in defined) or when a refresh is requested (XML analysis). For that a central storage for the modules configuration needs to be available to be parsed which are available for all ITmfTrace extensions. If a trace is opened a ITmfTrace instance is created and the applicable analysis module is instantiated and stored inside the ITmfTrace object.

Right now it's not possible to update available analysis module helpers in the TMF analysis framework based on a configuration specific for a trace instance.

This patch allows to add analysis modules to ITmfTrace and by-pass the TMF analysis framework and IAnalysisModuleSource extension point to provide custom analysis.

Note that those analysis modules won't show up in the Project Explorer when running Eclipse Trace Compass and hence it's meant for the trace server use case.

[Added] API to add and remove analysis modules to/from ITmfTrace

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #164, opened 2024-10-01
Customizable analysis 

bhufmann:
**Update (Oct 28, 0204):**
All the PRs below are merged. This PR only has on change left with some general workarounds applied in composite data providers left. This is needed for experiments that have traces where InstrumentedCallstackAnalysis are used and a trace doesn't have call stack events and it marks it as failed instead of completed with no call stack data.

-------------
This PR combines different commits to allow for customization of analysis using a json definition. These customable analyses are exposed through data providers used in conjunction with the Trace Compass trace server. 

Note that some of the commits are also available as separate pull requests.

List of commits/PRs:
- #141
- ~~#144~~
- #165
- #154
- #162
- #161
- #168 
- #166
- #171
- #172 tmf: Add data provider ProviderType.NONE for DPs without graphs
- WIP: Fix partial failed on composite data provider

Other modification not rebased in the PR
- #167 

### DIFF
--- a/analysis/org.eclipse.tracecompass.analysis.profiling.core/src/org/eclipse/tracecompass/internal/analysis/profiling/core/instrumented/FlameChartDataProviderFactory.java
+++ b/analysis/org.eclipse.tracecompass.analysis.profiling.core/src/org/eclipse/tracecompass/internal/analysis/profiling/core/instrumented/FlameChartDataProviderFactory.java
@@ -99,6 +99,7 @@ public class FlameChartDataProviderFactory implements IDataProviderFactory {
             if (!existingModules.contains(analysis.getId())) {
                 DataProviderDescriptor.Builder builder = new DataProviderDescriptor.Builder();
                 builder.setId(FlameChartDataProvider.ID + DataProviderConstants.ID_SEPARATOR + analysis.getId())
+                        .setParentId(analysis.getConfiguration() != null ? analysis.getId() : null)
                         .setName(Objects.requireNonNull(analysis.getName() + " - " +Messages.FlameChartDataProvider_Title)) //$NON-NLS-1$
                         .setDescription(Objects.requireNonNull(NLS.bind(Messages.FlameChartDataProvider_Description, analysis.getHelpText())))
                         .setProviderType(ProviderType.TIME_GRAPH)
--- a/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentNameSegmentStoreStatisticsDataProviderFactory.java
+++ b/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentNameSegmentStoreStatisticsDataProviderFactory.java
@@ -53,6 +53,7 @@ public class SegmentNameSegmentStoreStatisticsDataProviderFactory extends Abstra
     protected @Nullable IDataProviderDescriptor getDataProviderDescriptor(IAnalysisModule analysis) {
         DataProviderDescriptor.Builder builder = new DataProviderDescriptor.Builder();
         builder.setId(SegmentStoreStatisticsDataProvider.ID + DataProviderConstants.ID_SEPARATOR + analysis.getId())
+                .setParentId(analysis.getConfiguration() != null ? analysis.getId() : null)
                 .setName(Objects.requireNonNull(NLS.bind(Messages.SegmentStoreStatisticsDataProvider_title, analysis.getName())))
                 .setDescription(Objects.requireNonNull(NLS.bind(Messages.SegmentStoreStatisticsDataProvider_description, analysis.getHelpText())))
                 .setProviderType(ProviderType.DATA_TREE)
--- a/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreScatterDataProviderFactory.java
+++ b/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreScatterDataProviderFactory.java
@@ -73,6 +73,7 @@ public class SegmentStoreScatterDataProviderFactory implements IDataProviderFact
             if (!existingModules.contains(analysis.getId())) {
                 DataProviderDescriptor.Builder builder = new DataProviderDescriptor.Builder();
                 builder.setId(SegmentStoreScatterDataProvider.ID + DataProviderConstants.ID_SEPARATOR + analysis.getId())
+                    .setParentId(analysis.getConfiguration() != null ? analysis.getId() : null)
                     .setName(Objects.requireNonNull(NLS.bind(Messages.SegmentStoreScatterGraphDataProvider_title, analysis.getName())))
                     .setDescription(Objects.requireNonNull(NLS.bind(Messages.SegmentStoreScatterGraphDataProvider_description, analysis.getHelpText())))
                     .setProviderType(ProviderType.TREE_TIME_XY)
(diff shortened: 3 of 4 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/TmfJsonConfiguration.java
```diff
@@ -0,0 +1,294 @@
...
+    public static ITmfConfiguration fromJsonString(String json, ITmfConfiguration defaultValues) throws TmfConfigurationException {
+       try {
+           @SuppressWarnings("null")
+           TmfJsonConfiguration config = new Gson().fromJson(json, TmfJsonConfiguration.class);
+           TmfConfiguration.Builder builder = new TmfConfiguration.Builder();
+           String name = config.getName();
+           name = name == null ? defaultValues.getName() : name;
+           String description = config.getDescription();
+           description = description == null ? defaultValues.getDescription() : description;
+           return builder.setId(config.getId())
```
PatrickTasse, line 221: It seems a bit odd that we create a TmfConfiguration with id, name, description that can come from the json string and then the TmfConfiguration also contains the json string, so these values are duplicated inside the TmfConfiguration object.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -231,6 +232,38 @@ public List<IDataProviderDescriptor> getAvailableProviders(@Nullable ITmfTrace t
...
+     *            queried trace
+     * @param configId
+     *            a configuration ID used for the data provider
+     * @param ids
+     *            A list of ID of the data provider factory or a data provider ID of
+     *            form factoryId:secondaryId
+     * @return list of the available providers for this trace / experiment
+     * @since 9.5
+     */
+    public synchronized @NonNull List<IDataProviderDescriptor> getAvailableProviders(ITmfTrace trace, String configId, String... ids) {
```
abhinava-ericsson, line 249: For supporting per base DataProvider conifgurations We would need a `getAvailableProviders(ITmfTrace trace, String... ids)` to return all DataProviders which contain the (factoryId == baseDataProvider ID).
That, or DataProvider Descriptor object would need to be a tree of descriptor strings.

bhufmann: That could be useful. It would return all derived data providers for all configurations.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/trace/ITmfTrace.java
```diff
@@ -273,6 +273,37 @@ public interface ITmfTrace extends ITmfEventProvider {
      */
     @NonNull Iterable<@NonNull IAnalysisModule> getAnalysisModules();
 
+
```
bhufmann, line 276: @PatrickTasse @abhinava-ericsson 

This API needs explanation. In Trace Compass analysis modules are managed by the TmfAnalysisFramework through IAnalysisModuleSource extension. The ITmfTrace has a method getAnalysisModules() and that collection has only analysis modules that are configured through the analysis framework when opening a trace TmfTraceOpenSignal() processed which triggers the method executeAnalysis(). This then call TmfAnalysisManager.getAnalysisModules() which will then instantiate the modules and store in the trace object

Many data provider factories often use `TmfTraceUtils.getAnalysisModulesOfClass(trace, IFlameChartProvider.class);` which call `ITmfTrace.getAnalysisModules()`. So, if an analysis module is not know by the analysis framework, then analysis module is not instantiated and not return by the trace.

These 2 methods will by-pass the analysis framework and custom analysis modules can be added and removed. This is needed for the custom analysis implementation.

Please note that the analysis framework, reads some global configuration to instantiate so-called IAnalysisModuleHelpers which globally available and assigned to applicable traces. These helpers are used to instantiate the actual analysis module and then configure it with parameters provided by the extension point or xml definition.

One thing to remember, if we would like to use the analysis framework for custom analysis, which is doable and I prototyped, there needs to global storage of configuration files (e.g. json), a manager class that can add and remove such configurations as well as assign it to a trace. However, the design of customizale analysis through data providers (factories) doesn't support having such manager class and the required server endpoint.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/IAnalysisModule.java
```diff
@@ -138,6 +139,34 @@ public interface IAnalysisModule extends ITmfComponent, IAnalysisRequirementProv
      */
     @Nullable Object getParameter(@NonNull String name);
 
+    /**
+     * @param configuration
+     *            A ITmfConfiguration
+     * @return true if it was successful or false if invalid config (TODO should we through exception?)
+     * @since 9.5
+     */
+    default boolean setConfiguration(ITmfConfiguration configuration) {
```
bhufmann, line 148: This use needed to be able to add the configuration IDataProviderDescriptors for derived data providers. This is not used by the analysis framework, and hence doesn't have equivalent methods in IAnalysisModuleHelper.

@PatrickTasse @abhinava-ericsson

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/IAnalysisModule.java
```diff
@@ -138,6 +139,34 @@ public interface IAnalysisModule extends ITmfComponent, IAnalysisRequirementProv
...
+        return false;
+    }
+
+    /**
+     * @param configuration
+     *            A ITmfConfiguration
+     * @return true if it was successful or false if invalid config (TODO should we through exception?)
+     * @since 9.5
+     */
+    default @Nullable ITmfConfiguration getConfiguration() {
```
bhufmann, line 158: This use needed to be able to add the configuration IDataProviderDescriptors for derived data providers. This is not used by the analysis framework.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/IAnalysisModule.java
```diff
@@ -138,6 +139,34 @@ public interface IAnalysisModule extends ITmfComponent, IAnalysisRequirementProv
...
+        return null;
+    }
+
+    /**
+     * @return the configuration root path
+     * @since 9.5
+     */
+    default @Nullable String getConfigurationRoot() {
```
bhufmann, line 166: This can be removed. It's an artifact from a protoype that uses the anlaysis framework.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderDescriptor.java
```diff
@@ -87,4 +94,25 @@ public enum ProviderType {
...
     String getDescription();
+
+    /**
+     * Gets the parent data provider ID for grouping purposes.
+     *
+     * @return parent ID or null if not grouped or derived data provider
+     * @since 9.5
+     */
+    default @Nullable String getParentId() {
+        return null;
```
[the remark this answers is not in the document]
abhinava-ericsson: That sounds right. Currently, DataProviderManager maintains a list of all Data Providers, and we don't have nesting of derived data providers. So, base data providers don't really need to be aware of their children.
As you said, after full-stack implementation, if we see any performance or simplicity benefit by adding getChildren(), we can implement it.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderDescriptor.java
```diff
@@ -87,4 +94,25 @@ public enum ProviderType {
...
+    }
+
+    /**
+     * Gets the input configuration used to create this data provider.
+     *
+     * @return the {@link ITmfConfiguration} configuration use to create this
+     * data provider, or null if not applicable
+     * @since 9.5
+     */
+    default @Nullable ITmfConfiguration getCreationConfiguration() {
```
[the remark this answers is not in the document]
abhinava-ericsson: Just `getConfiguration()` is fine till we don't allow modification of an existing configuration. Later, if we do allow that, we can differentiate between `getConfiguration()` and `getInitialConfiguration()`.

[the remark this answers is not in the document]
bhufmann: `getConfiguration()` sounds like there can be only one. We might (I don't have use case yet) some other configuration, that does something else than creating a data provider. That's why I used `creationConfiguration`, but I don't like it. Maybe for now we call it `getConfiguration()` as you suggested and that's the configuration that was used for creation. If we in the future support another configuration we can find a name for the new type configuration.

PR #168, opened 2024-10-21
tmf: Allow TmfConfiguration to be de-/serialized with GSON

bhufmann:
Also, generate unique UUID as ID from name.

[Added] Allow TmfConfiguration to be de-/serialized with GSON 
[Added] TmfConfiguration.UNKNOWN string if values are not set [Added[ Utility methods to write/read configuration to/from JSON file

Continued from #115

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/TmfConfiguration.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/TmfConfiguration.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Ericsson
+ * Copyright (c) 2023, 2024 Ericsson
  *
  * All rights reserved. This program and the accompanying materials are
  * made available under the terms of the Eclipse Public License 2.0 which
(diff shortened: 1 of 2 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/TmfConfiguration.java
```diff
@@ -184,15 +290,79 @@ public Builder setParameters(Map<String, Object> parameters) {
...
-            if (id.isBlank()) {
-                throw new IllegalStateException("Configuration ID not set"); //$NON-NLS-1$
+            if (fId.isBlank()) {
+                fId = toUuidString(fName.isBlank() ? UNKNOWN : fName);
             }
             return new TmfConfiguration(this);
         }
     }
+    /**
+     * Converts a JSON parameters string to {@link ITmfConfiguration}
```
PatrickTasse, line 300: The input is not a string. But I wonder if this method is still needed?

PatrickTasse (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/TmfConfiguration.java
bhufmann: Yes, it's not needed anymore. I'll remove it.

PatrickTasse (approved):

PR #171, opened 2024-10-22
timing: set parentId and creation configuration in DataProviderDescriptor for Latency Analysis with conifgurations

bhufmann:
If a latency analysis was configured using an ITmfConfiguration instance, then set the configuration and parent ID in the Data Provider Descriptors of the corresponding data providers that being spawn by the Latency Analysis (e.g. Latency Statistics, Scatter Chart, Table and Flame Chart). This can be used for to show this information to end users of the trace server front-ends and for grouping of data providers in the front-end implementations.

- analysis: Set creation configuration to Latency Analysis DataProviderDescriptors
- timing: Use analysis id as parent ID for latency data providers with configurations. 

[Updated] Latency Analysis DataProviderDescriptors with config
[Updated] Latency Analysis DataProviderDescriptors with parentId


bhufmann:
CI is failing and needs to be fixed

PatrickTasse (approved):

PR #174, opened 2024-10-23
releng: Upgrade to Tycho 4.0.9

PatrickTasse:
Required to fix build failures related to org.eclipse.equinox.launcher in 4.34.

PR #176, opened 2024-10-24
tmf: remove unnecessary cast

arfio:
This patch remove some casts that are showing as errors in eclipse.

I found two categories:
- Cast from children classes to parent classes: I removed those ones since they are not really needed.
- Cast after an instanceof check: It seems that in some recent version, the behaviour changed and these ones are sometimes classified as unnecessary. I found information on JEP 305 where that allows the following:
```java
if (animal instanceof Cat cat)  {  
      out.println(cat.meow());  
}
```
However, in this case the following code indicates the cast as unnecessary:
```java
IResource resource = getResource();
if (resource instanceof IFolder) {
    path = ((IFolder) resource).getFullPath();
}
```
IResource being a parent class of IFolder.
As I am not sure what really happens there, I have not made the change to remove those casts even though removing it seems to work.


bhufmann:
Does this work with Java 11? We still support older Eclipse targets (e.g. 2021-06, e4.20) that builds with Java 11.

arfio:
I seem to have no errors building this with java 11 and eclipse 2021-06, e4.20. However I have a couple of errors with this PR: https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/174 for the changes added to the CompositeHostModel.

bhufmann (approved):
Looks good to me. Thanks.

MatthewKhouzam (approved):
Less code is a yes

PR #179, opened 2024-11-18
linux: support columns in ThreadStatusDataProvider tree model

bhufmann:
Do not merge until https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/1149 is fixed.


MatthewKhouzam (approved):
I was hesitant about exec, but it's everywhere in the kernel documentation.

MatthewKhouzam:
We can call exec process. If and only if you want.

bhufmann:
> We can call exec process. If and only if you want.

The string `Exec_name` is used as label in other analysis and views (e.g. `Scheuler Wakeup To Scheduler Switch Latency - Latency Table`). I think it should be consistent across views.

image

MatthewKhouzam:
Approved!

PR #181, opened 2024-11-19
tmf.ui: instrument histogram

MatthewKhouzam:
Add early exit when the model is unchanged

Change-Id: I8e89fa289e071a56220d410215c3ed84c5b866c1


bhufmann (changes requested):
This PR does more than instrument the histogram. It also adds some update to the code. The instrumentation should be a separate commit and the code change in another commit. The instrumentation should come first. you can both commits in this PR.

bhufmann (approved):
LGTM. Thanks!

PR #188, opened 2024-11-25
releng: Update e4.34 and eStaging targets for 2024-12 RC2

bhufmann:


MatthewKhouzam (approved):
We're out version 240!

PR #196, opened 2024-12-09
json: Add an offset to sorting job

MatthewKhouzam:
The json trace files can have events before the "traceevents": key. This patch allows the parser to start sorting on the correct entry.

This allows traces generated by pytorch to be read by the incubator as they add out-of-spec metadata information.

Example trace generated: 
[stable-diffusion.json](https://github.com/user-attachments/files/18063112/stable-diffusion.json)

Example of how the trace can help accelerate pytorch: 
https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/716

Devs can see the impact of the arguments they are passing to the API.

This patch needs an incubator counterpart.


arfio (commented):
Should the processMetadata method should be used instead of just skipping the bytes ? I do not think that adding an additional constructor is warranted here whereas there is already a mechanism in place that should work if the metadata is before or after the events in the json format.

MatthewKhouzam:
We still need to skip forward, would having processmetadata return a start offset be OK in your opinion? I am trying, I like your approach but I think the caffeine is failing me.

arfio:
> We still need to skip forward, would having processmetadata return a start offset be OK in your opinion?
process metadata is called at the end of events, but I am not sure what the reasoning was when this was written. I think it should be called before reading and sorting the events, if the metadata is at the end of the file, then it is the responsibility of the implementation to find and read the metadata information.
Or it could be a setting where the implementation can return if the metadata is at the beginning or at the end.
In any case the implementation has access to the BufferedInputStream and can skip the bytes itself.

PR #197, opened 2024-12-09
profiling: cache symbol key interval

arfio:
To resolve symbols, the method retrieveSymbolKeyAt will query a symbol key attribute for every interval in a query2d. This creates a significant overhead that can easily be avoided if we cache this interval. For all current use cases, the symbol key interval is unique for each trace.


MatthewKhouzam (approved):
Looks great, can I have a trace before and after? I would love to play with execution comparisons.

bhufmann (approved):

PR #206, opened 2025-01-22
Add json to the category.xml

merks:
### What it does

While investigating disabling ECF's SimRel contribution I discovered that tracecompass requires json and ECF is the only update site that provide json.

### How to test

This only affects the content of the update site.

I have verified by building the update site locally that SimRel can aggregate without ECF's udpate site contribution.

image

(I needed to address Mylyn separately).

### Review checklist


bhufmann (approved):
Looks good to me. Thanks!

PatrickTasse (approved):

PR #207, opened 2025-01-23
common: Deprecate TraceCompassLogUtils and use trace-event-logger lib

PatrickTasse:
### What it does

Deprecate TraceCompassLogUtils and TraceCompassMonitorManager.

Replace all use of TraceCompassLogUtils with LogUtils from the
trace-event-logger library.

Update all targets to include the trace-event-logger jar from Maven.

Add trace-event-logger plug-in to RCP feature and releng-site.

### How to test

Unit tests.

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -44,7 +44,12 @@ Trace Compass can be traced by doing the following in the launch configuration:
 * -Djava.util.logging.config.file=%gitroot%/logging.properties (where %gitroot% is the directory tracecompass is checked out to)
 * -Dorg.eclipse.tracecompass.logging=true
 
-Additionally the folder the trace is being written to (default is `home/.tracecompass/logs`) needs to be created in advance. After running Trace Compass, a `trace_n.json` will be created in the tracing folder. It needs to be converted to true json, so use the `jsonify.sh` script in the root directory to convert it. Then it can be loaded into Trace Compass, if the **Trace Event format** is installed from the incubator, or from a web browser such as Chromium.
+Additionally the folder the trace is being written to (default is `%home%/.tracecompass/logs`) needs to be created in advance. After running Trace Compass, a `trace_n.json` will be created in the tracing folder. It needs to be converted to true json, so use the `jsonify.sh` script in the root directory to convert it. Then it can be loaded into Trace Compass, if the **Trace Event format** is installed from the incubator, or from a web browser such as Chromium.
+
+To use the handlers provided in the trace-event-logger library by adding them to the logging config file, they will need to be added to the bootstrap class path in vmargs using a relative path (RCP) or absolute path (development):
+
+* -Xbootclasspath/a:./plugins/org.eclipse.tracecompass.trace-event-logger_x.y.z.jar
+* -Xbootclasspath/a:%home%/.m2/repository/org/eclipse/tracecompass/trace-event-logger/x.y.z/trace-event-logger-x.y.z.jar
 
 The performance impact is low enough as long as the log level is greater than "*FINEST*".
 
--- a/common/org.eclipse.tracecompass.common.core/src/org/eclipse/tracecompass/common/core/log/TraceCompassLogUtils.java
+++ b/common/org.eclipse.tracecompass.common.core/src/org/eclipse/tracecompass/common/core/log/TraceCompassLogUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Ericsson
+ * Copyright (c) 2016, 2025 Ericsson
  *
  * All rights reserved. This program and the accompanying materials are
  * made available under the terms of the Eclipse Public License 2.0 which
@@ -25,6 +25,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.tracecompass.traceeventlogger.LogUtils;
 
 /**
  * Logger helper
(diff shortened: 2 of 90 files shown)


common/org.eclipse.tracecompass.common.core/src/org/eclipse/tracecompass/common/core/log/TraceCompassLogUtils.java
```diff
@@ -255,37 +175,31 @@ public static class FlowScopeLogBuilder {
...
+            wrapped.setCategory(category);
             return this;
         }
 
         /**
          * Set a category and ID for the flow scope.
          *
          * This method is mutually exclusive with
-         * {@link #setParentScope(FlowScopeLog)}. Calling both will throw an
+         * {@link #setParentScope(IFlowScopeLog)}. Calling both will throw an
```
bhufmann, line 202: This needs to be change to FlowScopeLog

common/org.eclipse.tracecompass.common.core/src/org/eclipse/tracecompass/common/core/log/TraceCompassLogUtils.java
```diff
@@ -255,37 +175,31 @@ public static class FlowScopeLogBuilder {
...
+            wrapped = new LogUtils.FlowScopeLogBuilder(logger, level, label, args);
         }
 
         /**
          * Set a category for the flow scope. When building the scope, an ID
          * will be automatically generated.
          *
          * This method is mutually exclusive with
-         * {@link #setParentScope(FlowScopeLog)}. Calling both will throw an
+         * {@link #setParentScope(IFlowScopeLog)}. Calling both will throw an
```
bhufmann, line 186: This needs to be change to FlowScopeLog

bhufmann (changes requested):

bhufmann (changes requested):
I did a first pass. Please rebase this PR. The jsonify.sh script doesn't work anymore. Maybe copy the python script of the trace-event-logger repository.

With the deprecated logger, it was necessary to set -Dorg.eclipse.tracecompass.logging=true. The TraceCompassLogUtils class had code to disable the logging if that flag is not set. That's to avoid unnecessary logging due to unrelated logging.properties. We need to decide if we need this and where to set if if needed, either trace-event-logger level or in Trace Compass repo only.

The BUILDING.md file needs to be updated for the new usage of the library. Also, the logging.properties that is in the root folder of the repo needs to be updated. Mabye, we should have a couple examples with typical settings for different handlers including the FileHandler.

To make the AsyncFileHandler and SnapshotHandler work, you need to set the bootclasspath. From the RCP you can use the relative path. From the IDE development environment you will have to provide an absolute path.

```
-Xbootclasspath/a:./plugins/org.eclipse.tracecompass.trace-event-logger_0.2.0.jar
-Djava.util.logging.config.file=goodlogging.properties
-Dorg.eclipse.tracecompass.logging=true
```

PatrickTasse:
> I did a first pass. Please rebase this PR. The jsonify.sh script doesn't work anymore. Maybe copy the python script of the trace-event-logger repository.
> 
> With the deprecated logger, it was necessary to set -Dorg.eclipse.tracecompass.logging=true. The TraceCompassLogUtils class had code to disable the logging if that flag is not set. That's to avoid unnecessary logging due to unrelated logging.properties. We need to decide if we need this and where to set it if needed, either trace-event-logger level or in Trace Compass repo only.
> 
> The BUILDING.md file needs to be updated for the new usage of the library. Also, the logging.properties that is in the root folder of the repo needs to be updated. Mabye, we should have a couple examples with typical settings for different handlers including the FileHandler.
> 
> To make the AsyncFileHandler and SnapshotHandler work, you need to set the bootclasspath. From the RCP you can use the relative path. From the IDE development environment you will have to provide an absolute path.
> 
> ```
> -Xbootclasspath/a:./plugins/org.eclipse.tracecompass.trace-event-logger_0.2.0.jar
> -Djava.util.logging.config.file=goodlogging.properties
> -Dorg.eclipse.tracecompass.logging=true
> ```

The commit is rebased.

The bash script jsonify.sh is changed to be a wrapper to download and execute jsonify.py from trace-event-logger repository. The downloaded script is added to .gitignore file.

The property -Dorg.eclipse.tracecompass.logging=true is still needed as it activates the Logger instances that are passed to the trace-event-logger library.

BUILDING.md
```diff
@@ -44,7 +44,7 @@ Trace Compass can be traced by doing the following in the launch configuration:
 * -Djava.util.logging.config.file=%gitroot%/logging.properties (where %gitroot% is the directory tracecompass is checked out to)
 * -Dorg.eclipse.tracecompass.logging=true
```
bhufmann, line 45: Should we add some description here on how to use the other handlers and how to make it work with the `-Xbootclasspath` setting? We should also have a reference link to the trace-event-logger documentation for more information.

The README.md in this repo should mention about tracing Trace Compass and have link to the chapter in BUILDING.md.

bhufmann (changes requested):
I was able to create a trace and load it into Trace Compass. Some small comments though.

bhufmann (changes requested):
Since the trace event logger v0.3.0. supports MXBeans, TraceCompassMonitorManager and related code can be deprecated as well

bhufmann (approved):
Looks good to me. I tested it by creating a trace on my laptop with a local build. Thanks!

PR #233, opened 2025-03-12
tmf: Support hideOutput extension point element for data providers

PatrickTasse:
### What it does

Define constant DataProviderConstants.CONFIG_SEPARATOR.

Describe use of CONFIG_SEPARATOR in ITmfDataProviderConfigurator's createDataProviderDescriptors() method.

In DataProviderManager.getAvailableProviders(), filter-out data provider descriptors that are hidden by org.eclipse.linuxtools.tmf.core.analysis extension point's hideOutput element, ignoring any configuration id suffix. Extract the analysisId from the provider id's secondary id.

Describe use of data provider id as possible value of org.eclipse.linuxtools.tmf.core.analysis extension point's hideOutput element's output attribute.

Change org.eclipse.linuxtools.tmf.core.analysis extension point's hideOutput element's analysis_module attribute to be optional. This will enable hiding of data provider configurators.

### How to test

Define data provider outputId's in the hideOutput extension point element.

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 Ericsson
+ * Copyright (c) 2017, 2025 Ericsson
  *
  * All rights reserved. This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0 which
@@ -26,6 +26,7 @@ import org.eclipse.core.runtime.Platform;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.tracecompass.internal.tmf.core.Activator;
+import org.eclipse.tracecompass.tmf.core.analysis.TmfAnalysisOutputManager;
 import org.eclipse.tracecompass.tmf.core.component.DataProviderConstants;
 import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataModel;
 import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataProvider;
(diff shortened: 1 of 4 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
```diff
@@ -222,10 +224,23 @@ public List<IDataProviderDescriptor> getAvailableProviders(@Nullable ITmfTrace t
             return Collections.emptyList();
         }
         List<IDataProviderDescriptor> list = new ArrayList<>();
+        TmfAnalysisOutputManager manager = TmfAnalysisOutputManager.getInstance();
```
bhufmann, line 227: This introduces a coupling of analysis modules and data providers. Data Providers don't need to use analysis modules. However, in most case so far there is analysis module defined that the data provider is using. But, if we look outside the code coupling, data providers are outputs and hence it probably ok to expand the hideOutput extension point to data providers, especially, since analysis_id is optional in the extension point.

bhufmann (changes requested):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/DataProviderManager.java
PatrickTasse: It's not adding dependencies on modules however it causes the data provider manager to need to consider analysisId inside the outputId. We're only accessing the TmfAnalysisOutputManager to reuse its reading of the hideOutput extension point attribute. This could be factored out, however it would still be tied to the 'org.eclipse.linuxtools.tmf.core.analysis' extension point.

Perhaps it would be better to replicate the functionality of 'hideOutput' in 'org.eclipse.tracecompass.tmf.core.dataprovider' extension point? Then we can remove the handling of analysisId completely (it could still be part of the outputId string but that would be opaque to the implementation).

bhufmann: Sounds reasonable to have it in the data provider extension point. Would you implement a regex pattern to be able hide all data providers with for example the same customization?

PatrickTasse:
Replaced with https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/243

PR #235, opened 2025-03-13
Propose new api for configurable data provider factories

frallax:
### What it does

Proposal to extend the public API with a new class that can (should) be extended by all data provider factories that want to provide the *configuration* feature.

The main goal is to provide a "pattern" for configurable data provider factories, so that they follow a similar coding style (at least for the less advanced use cases). This hopefully can help to make configurable data provider factories easier to implement and to document how to implement them.

A practical example of how to use this is provided in a separate PR for the InAndOutDataProvider in the incubator project (I will add comments below).

### How to test

Tests will be added if the proposal is accepted.
For the moment, checking out this branch together with the one named similarly for the incubator project, where the InAndOutDataProvider is used, it should enable seeing how the code changes while the InAndOutDataProvider has the same functions as before.

### Follow-ups

If accepted, adding tests and documentation on how to create configurable data provider factories using this common pattern.

### Review checklist


frallax:
NOTES:
- as a first draft, I am pushing to 2021-06 just because I have a environment configured on that env. It will eventually be pushed to master if discussions go in the right direction
- please ignore changes to non-java files for the draft PR
- I will add a link to incubator PR that shows how I foresee the usage of the new API

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
```diff
@@ -0,0 +1,242 @@
...
+        if (configuration.getSourceTypeId().equals(TmfConfiguration.UNKNOWN)) {
+            throw new TmfConfigurationException("Missing configuration type"); //$NON-NLS-1$
+        }
+
+        String description = configuration.getDescription();
+        if (configuration.getDescription().equals(TmfConfiguration.UNKNOWN)) {
+            description = "Data provider defined by configuration " + configuration.getName(); //$NON-NLS-1$
+        }
+
+        TmfConfiguration.Builder builder = new TmfConfiguration.Builder();
```
siwei-zhang-work, line 64: This class currently enforces the use of `TmfConfiguration`, A better approach might be to introduce a `createConfig()` method, which child classes can implement to return an `ITmfConfiguration`.

```
    @Override
    protected ITmfConfiguration createConfig() {
        return new CustomTmfConfiguration();
    }
```

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
```diff
@@ -0,0 +1,242 @@
...
+            return;
+        }
+        try {
+            if (trace instanceof TmfExperiment) {
+                for (ITmfTrace tr : TmfTraceManager.getTraceSet(trace)) {
+                   // Read configurations from sub-trace
+                   List<ITmfConfiguration> configs = readConfigurations(tr);
+                   readAndApplyConfiguration(trace, configs);
+                }
+            } else {
```
siwei-zhang-work, line 150: The configuration might be stored at the experiment level, so it might be better to move the following two lines outside the else block to ensure they always execute.

```
List<ITmfConfiguration> configs = readConfigurations(trace);
readAndApplyConfiguration(trace, configs);
```

bhufmann (changes requested):
Thanks for the contribution. It's good to have code re-use. The incubator InAndOutAnalysis and other downstream plug-ins with similar configurations can use it. I added some comments this PR.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
frallax: tried to fix, check if you agree

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
frallax: tried to fix, check if you agree

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
siwei-zhang-work: Done, thanks.

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/config/AbstractTmfDataProviderConfigurator.java
```diff
@@ -0,0 +1,242 @@
...
+     * @since 9.5
+     */
+    public static final String JSON_EXTENSION = "json"; //$NON-NLS-1$
+
+    private Table<String, ITmfTrace, ITmfConfiguration> fTmfConfigurationTable = HashBasedTable.create();
+
+    /**
+     * @return a table mapping configuration id and trace (exp) to its configuration
+     */
+    protected Table<String, ITmfTrace, ITmfConfiguration> getConfigurationTable(){
```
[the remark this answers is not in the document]
frallax: I have not understood what you propose here. Could you please try to rephrase or provide some code snippets? Thanks!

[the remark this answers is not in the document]
bhufmann: So, the `IDataProviderFactory` provides a `getAdapter(class)` method (https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/a081aac03c2846d1f9d97831a06fb9a66a57f776/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/dataprovider/IDataProviderFactory.java#L97) that the trace server is using to get the configurator instance. The factory implementation can decide to implement itself the `ITmfDataProviderConfigurator` interface or overrides the `getAdapter(...)` method as below and return a different class that implements the `ITmfDataProviderConfigurator` interface. So, the `getConfigurationTable()` method needs to be public instead of protected to allow that the factory class to call that method.

```java
// ....
   private MyConfigurator fConfigurator = new MyConfigurator();

   @Override
    public <T> @Nullable T getAdapter(@Nullable Class<T> adapter) {
        if (adapter != null && adapter.equals(ITmfDataProviderConfigurator.class)) {
            return (T) fConfigurator;
        }
        return null;
    }

// ...
```

where:

```java
public class MyConfigurator implements ITmfDataProviderConfigurator {
// ...
}
```

So, if `MyConfigurator` would extend your new `AbstractTmfDataProviderConfigurator`  the factory class might call `fConfigurator.getConfigurationTable()` to get configurations per trace. Here is the change that I had to do for the InAndOutAnalysis method getDescriptor():

``java
@Override
    public Collection<IDataProviderDescriptor> getDescriptors(ITmfTrace trace) {
        List<IDataProviderDescriptor> list = new ArrayList<>();
        list.add(DESCRIPTOR);
        for (ITmfConfiguration config : configurator.getConfigurationTable().column(trace).values()) {
            list.add(fConfigurator.getDescriptorFromConfig(config));
        }
        return list;
    }
```

Having said all the above, you might be able to change the method configurator.getDescriptorFromConfig(...) and pass the trace as parameter. Then you don't need to make the getConfigurationTable() public.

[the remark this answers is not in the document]
frallax: I made the method public as you requested

PR #243, opened 2025-03-26
tmf: Add hideDataProvider extension point element

PatrickTasse:
### What it does

Add hideDataProvider element with idRegex and tracetype attributes to
the org.eclipse.tracecompass.tmf.core.dataprovider extension point.

Add dataprovider.ini configuration file template to be stored in the
o.e.tc.tmf.core plug-in state location.

In DataProviderManager methods, filter-out data provider descriptors and
factories that are hidden by the hideDataProvider element, or the
dataprovider.ini configuration file, but that are not shown by the
configuration file. Note that factories cannot be hidden for a specific
trace type.

Define constant DataProviderConstants.CONFIG_SEPARATOR.

Describe use of CONFIG_SEPARATOR in ITmfDataProviderConfigurator's
createDataProviderDescriptors() method.

### How to test

Add hideDataProvider element in org.eclipse.tracecompass.tmf.core.dataprovider extension to any plugin.xml, with idRegex for a specific or all trace types. Check matching data providers are not returned by getAvailableProviders().

Add hide or show settings to <workspace>/.metadata/.plugins/org.eclipse.tracecompass.tmf.core/dataprovider.ini. Check hide matching data providers are not returned, and show matching data providers are returned despite being hidden by hideDataProvider element.

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/releng/org.eclipse.tracecompass.integration.core.tests/build.properties
+++ b/releng/org.eclipse.tracecompass.integration.core.tests/build.properties
@@ -14,4 +14,5 @@ bin.includes = META-INF/,\
                .,\
                plugin.properties,\
                about.html,\
-               build.properties
+               build.properties,\
+               plugin.xml
--- a/releng/org.eclipse.tracecompass.integration.core.tests/plugin.xml
+++ b/releng/org.eclipse.tracecompass.integration.core.tests/plugin.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="org.eclipse.tracecompass.tmf.core.dataprovider">
+      <dataProviderFactory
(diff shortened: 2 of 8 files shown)


bhufmann (changes requested):
I started testing the PR. I'm not fully through with testing and review, but I'd like to give some initial feedback. I noticed that when hiding a data provider ( e.g.  `org.eclipse.tracecompass.analysis.os.linux.core.cpuusage.CpuUsageDataProvider`) from the plugin.xml, it's not showing up in the available data providers, but it's still to request the XY tree and hence the data provider is still created. I think it should not be possible to create the data providers when hidden.

Could you please also some unit tests for that.

bhufmann (approved):
Looks good to me. I tested many different use cases and pattern and it worked well. I realized that when adding a specific ID that has `.` in it, that it technically means any characters because it's a regular expression. Users should escape it, i.e. `\.` to be sure the regex doesn't match other data providers accidentally.

Thanks for this contribution.

PR #246, opened 2025-04-02
Use the current size in segment store 

siwei-zhang-work:
### What it does

Set the size of the segment store even if the provider doesn't implement `IAnalysisModule`. This change was made to support cases where `SegmentStoreTableDataProvider` is used with classes that do not implement `IAnalysisModule`.

Before this change:

Only analysis modules implementing `IAnalysisModule` will have their segment count recorded. Otherwise, the segment count will be treated as 0, resulting in no rows being displayed in the virtual table.

### How to test

Everything should work the same as before. Only the possibility to use other classes is added.

### Follow-ups

No follow-ups.

### Review checklist

As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

### DIFF
--- a/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreTableDataProvider.java
+++ b/analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreTableDataProvider.java
@@ -327,9 +327,9 @@ public class SegmentStoreTableDataProvider extends AbstractTmfTableDataProvider 
     public TmfModelResponse<TmfTreeModel<TmfTreeDataModel>> fetchTree(Map<String, Object> fetchParameters, @Nullable IProgressMonitor monitor) {
         if (fSegmentProvider instanceof IAnalysisModule) {
             ((IAnalysisModule) fSegmentProvider).waitForCompletion();
-            ISegmentStore<ISegment> segmentStore = fSegmentProvider.getSegmentStore();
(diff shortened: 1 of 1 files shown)


analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreTableDataProvider.java
```diff
@@ -389,7 +386,7 @@ public TmfModelResponse<ITmfVirtualTableModel<VirtualTableLine>> fetchLines(Map<
             if (segStore.isEmpty()) {
                 return new TmfModelResponse<>(new TmfVirtualTableModel<>(columnIds, Collections.emptyList(), queryFilter.getIndex(), 0), ITmfResponse.Status.COMPLETED, CommonStatusMessage.COMPLETED);
             }
-            if (queryFilter.getIndex() >= fSegmentStoreSize) {
+            if (queryFilter.getIndex() >= segStore.size()) {
```
MatthewKhouzam, line 389: The reason we have the variable stored is to not recalculate the size at each iteration. Can you maybe check if fSegStoreSize != 0 at line 390?

We need to observe the performance impact of this patch.

siwei-zhang-work: I see that now that the segStore.size() can have a big performance impact. I changed the commit to only move the initialization out of the if block.

MatthewKhouzam (approved):
I like this a lot! Simple and clean!

bhufmann (approved):
Looks good to me. Thanks!

PR #251, opened 2025-04-09
Add data fetching capabilities to data providers

kavehshahedi:
### What it does

Introduce a new interface and model class to allow data providers to fetch and return custom data based on provider configuration.
- `ITmfDataProviderConfigurationDataFetcher` interface that allows data providers to implement custom data retrieval capabilities without requiring specific knowledge of the data format in the core API.
- `TmfDataProviderConfigurationDataModel` class to encapsulate different types of content (files, byte arrays, strings, etc.) along with associated metadata such as content type and filename.

This extension enables data providers to serve various types of content (reports, images, data exports, etc.) in a standardized way while keeping the core API independent of specific implementations.

### How to test

Run the project tests via `mvn clean test`.

### Follow-ups

Any data provider configuration that needs to have such a functionality can implement the `ITmfDataProviderConfigurationDataFetcher` interface and provide the required logic in order to retrieve its data.

### Review checklist


MatthewKhouzam (dismissed):

tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/dataprovider/TmfDataProviderConfigurationDataModel.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Get the content of the data model.
+     * The content type is Object as it can be anything (e.g. image, report, etc.)
+     *
+     * @return The content of the data model
+     */
+    public Object getContent() {
+        return fContent;
```
bhufmann, line 48: Object could be anything. As a extender I would not know what I would have to return here.

PR #280, opened 2025-06-10
Do not merge! Test flame graph data provider

bhufmann:
### What it does

Expose FlameGraphDataProvider using new data provider type. Implement `getDescriptor` method for FlameGraphDataProviderFactory so that it's visible when running the trace server.

### How to test
Query available data providers using Python client.

### Follow-ups

Note: TBD if we use this data provider implementation or create a weighted tree viewer data provider and clients can implement different ways to visualize that data: flame graph, data tree table, tree map. 

### Review checklist


MatthewKhouzam (approved):
Good catch!

PR #289, opened 2025-06-16
linux: set auto-expand-level in ResourcesStatusDataProvider

bhufmann:
Note: The classic Eclipse UI has been doing this in the UI code without that setting (see ResourcesView).

### What it does

linux: set auto-expand-level in ResourcesStatusDataProvider

Setting the auto-expand-level to 1 of the returned TmfTreeModel when calling fetchTree() will indicate to the clients that only the top-level should be expanded and their children supposed to be collapsed.

With this trace server clients can apply this new setting and it fixes FE issue https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/801.

### How to test

Run server with this change and run theia-trace-extension change https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1202.

### Follow-ups

N/A

### Review checklist


marcdumais-work (approved):
(placebo approval, as a non-committer) LGTM! 


image

PatrickTasse (approved):

PR #295, opened 2025-07-16
Mark `name` attributes for `org.eclipse.linuxtools.tmf.core` extension point as translatable

ruspl-afed:
Names of
* category
* type
* experiment 

are user-visible string and should have corresponding metadata to help PDE control its values.

### What it does

PDE doesn't trigger on non-externalized name attribute.
This PR adds "translatable" metadata for `name` attribute to the extension point schema 

### How to test

After applying this change (and restarting IDE) PDE should recognize mentioned attributes as translatable
(use "Plug-in Tools -> Externalize String" from the project context menu)

image

### Follow-ups

I would say that this PR introduces technical debt, it is rather will help to resolve existing one with not translated names.

### Review checklist


bhufmann (approved):
Looks good to me. Thanks for the contribution!

PatrickTasse (approved):

PR #296, opened 2025-07-22
Revert unnecessary tmf.core minor version bump

bhufmann:
### What it does

Revert unnecessary tmf.core minor version bump

The minor version has already been increased from 10.0.0 to 10.1.0 since
the previous Trace Compass baseline.

### How to test

Open source code in Eclipse with tracecompass-baseline-11.0.0.target set as API baseline. No API errors should be shown in the updated class.

### Follow-ups
N/A

### Review checklist


ruspl-afed:
I saw this and I wasn't sure if it was a problem with the baseline configuration or with the tag itself.

PatrickTasse (changes requested):
Plug-in o.e.tc.tmf.core version was already stepped from 10.0 to 10.1 by auto-expand PR, then it was stepped again by gantt chart PR. It should remain at 10.1 so these since tags are ok, it's the MANIFEST.MF version bump that should be reverted.

MatthewKhouzam (approved):

PatrickTasse (approved):

PR #304, opened 2025-08-20
tmf: Fix IAxisDomain toString() and Javadoc

PatrickTasse:
### What it does

Change Range.toString() to use "Range" instead of "TimeRange".

Change both toString() to use "IAxisDomain" instead of "AxisDomain".

Change Javadoc to refer to generic axis instead of only X axis.

### How to test

Select an AxisDomain variable in debug mode Variables view.

### Follow-ups

N/A

### Review checklist


bhufmann:
FYI @siwei-zhang-work

bhufmann (approved):
Looks good to me. Thanks!

PR #305, opened 2025-08-21
lttng: Make text metadata detection more lenient

### LINKED ISSUE #298: LTTng 2.14 trace not recognize
### Steps to Reproduce:

1.
2.
3.

### Additional Information

- Operating System:
- Trace Compass Version:

bhufmann:
### What it does
Text CTF metadata files are supposed to start with "/* CTF " followed by the CTF version. https://diamon.org/ctf/v1.8.3/#spec7.1

However, to support opening traces that don't comply with the specification check for some common metadata keywords at the beginning of the metadata file when validating the trace. This will allow to open such traces with Trace Compass.

fixes #298

### How to test
Try to open trace attached in #298 which should work now.

### Review checklist


MatthewKhouzam (approved):
Great, it should help the embedded world too. I would be tempted to add a warning if it doesn't start with /* CTF... but this works.

MatthewKhouzam (approved):

PR #310, opened 2025-08-29
tmf: Propagate analysis failure of dependent analysis to parent

### LINKED ISSUE #301: Failure on dependent analysis wouldn't fail parent analysis
### Bug Description

Failures in a dependent analysis don’t propagate to the parent analysis.

When a module fails (`TmfAbstractAnalysisModule#fail()`), its `waitForCompletion()` correctly returns false. However, the parent module does not check (or honor) the dependency’s `waitForCompletion()` result, so the parent proceeds as if the dependency succeeded and continues executing, returning partial results.

Steps to Reproduce

1. Create Module A that depends on Module B.

2. During Module B’s execution, throw an exception so that fail() is invoked.

3. Observe that Module A still runs and produces (partial) output, instead of failing or aborting.

Expected: If a dependency fails, the parent’s waitForCompletion() should reflect that failure (and the parent should not proceed).
Actual: The parent continues and behaves as if the dependency completed successfully.

### Additional Information

Current version: master

Place where if can be fixed (in `TmfAbstractAnalysisModule` line 395-397):
```
                            for (IAnalysisModule module : dependentAnalyses) {
                                module.waitForCompletion(mon);
                            }
```
`module.waitForCompletion(mon);` will return false if it failed, checking it there would work.

bhufmann:
### What it does

tmf: Propagate analysis failure of dependent analysis to parent

fixes #301

### How to test

Verify using added Junit test

### Follow-ups

None

### Review checklist

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/TmfAbstractAnalysisModule.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/TmfAbstractAnalysisModule.java
@@ -300,6 +300,14 @@ public abstract class TmfAbstractAnalysisModule extends TmfComponent
     }
 
     /**
+     * @since 10.1
+     */
+    @Override
+    public @Nullable Throwable getFailureCause() {
+        return fFailureCause;
+    }
+
+    /**
      * Method executed when the analysis has failed, so that analysis can
      * rectify their state. For instance, if the analysis had not been
      * initialized when the exception occurred, this method could mark the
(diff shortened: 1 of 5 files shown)


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/analysis/TmfAbstractAnalysisModule.java
```diff
@@ -383,7 +391,7 @@ private void execute(final ITmfTrace trace) {
              * Actual analysis will be run on a separate thread
              */
             String jobName = checkNotNull(NLS.bind(Messages.TmfAbstractAnalysisModule_RunningAnalysis, getName()));
-            fJob = new Job(jobName) {
+            Job job = new Job(jobName) {
```
bhufmann, line 394: No idea how this change happened

PatrickTasse (approved):

PR #312, opened 2025-09-02
timing: Return FAILED when analysis failed in SegmentStoreStatisticsDp 

### LINKED ISSUE #302: Segment store provider sends back partial results with COMPLETED status
### Bug Description:

In `SegmentStoreStatisticsDataProvider` and `SegmentStoreTreeStatisticsDataProvider`, when the execution of the analysis fails, it returns `COMPLETED` status.

### Steps to Reproduce:

1. Run an analysis with `SegmentStoreStatisticsDataProvider` and fail (TmfAbstractAnalysisModule#fail()) it in progress.
2. The data provider will send back partial result with `COMPLETED` status.

### Additional Information

- Trace Compass Version: master
- Root cause (in `SegmentStoreStatisticsDataProvider`):
```
        if (module != null) {
            if (monitor != null) {
                module.waitForCompletion(monitor);
                if (monitor.isCanceled()) {
                    return new TmfModelResponse<>(null, Status.CANCELLED, CommonStatusMessage.TASK_CANCELLED);
                }
            } else {
                module.waitForCompletion();
            }
        }

```

The status of waitForCompletion() can be checked, as it'll return false when the analysis fails.

bhufmann:
### What it does

timing: Return FAILED when analysis failed in SegmentStoreStatisticsDp

Fixes #302

### How to test

Verify with added JUnit test case (`SegmentStoreStatisticsDataProviderTest#testModuleExecutionError`)

### Follow-ups
N/A

### Review checklist


analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreStatisticsDataProvider.java
```diff
@@ -116,17 +117,26 @@ public SegmentStoreStatisticsDataProvider(ITmfTrace trace, IStatisticsAnalysis p
...
                 }
             } else {
-                module.waitForCompletion();
+                success = module.waitForCompletion();
+                if (!success && module.getFailureCause() == null) {
+                    return new TmfModelResponse<>(null, Status.CANCELLED, CommonStatusMessage.TASK_CANCELLED);
+                }
+            }
+
+            if (!success && module.getFailureCause() != null) {
```
PatrickTasse, line 138: What should we do in case of failure but the failure cause is null?

analysis/org.eclipse.tracecompass.analysis.timing.core/src/org/eclipse/tracecompass/internal/analysis/timing/core/segmentstore/SegmentStoreStatisticsDataProvider.java
```diff
@@ -116,17 +117,26 @@ public SegmentStoreStatisticsDataProvider(ITmfTrace trace, IStatisticsAnalysis p
...
+    @SuppressWarnings("null")
     @Override
     public TmfModelResponse<TmfTreeModel<SegmentStoreStatisticsModel>> fetchTree(Map<String, Object> fetchParameters, @Nullable IProgressMonitor monitor) {
         IAnalysisModule module = fModule;
+        boolean success = true;
         if (module != null) {
             if (monitor != null) {
-                module.waitForCompletion(monitor);
+                success = module.waitForCompletion(monitor);
                 if (monitor.isCanceled()) {
```
PatrickTasse, line 128: If there is failure but also monitor canceled, should the failure override the cancel? i.e. should we return cancelled only if there is no failure cause?

PatrickTasse (changes requested):

PatrickTasse (approved):

PR #324, opened 2025-09-17
tmf.ctf: Re-initialize start and endtime after transform is applied

### LINKED ISSUE #323: Properly offset time in ctf
### Bug Description:

The automatic trace offsetting when opening a kernel and ust trace from the same host as experiment is not correctly stored in the trace instance when opening the trace the first time. This is only really noticeable when using the trace server with the trace viewer and python client.

In the events table the endtime of the kernel trace is correctly shown after the offsetting is applied. However, the trace endtime of the kernel trace when fetching the experiment details in the python client is incorrect:
```
Successfully listed experiment
------------------------------
{
    "name": "context-switches",
    "UUID": "1605d743-89be-3197-92d7-8ac2d915fdd4",
    "start": 1450193684107249704,
    "end": 1450193776505076263,
    "nbEvents": 245500,
    "indexingStatus": "COMPLETED",
    "traces": [
        {
            "name": "context-switches-kernel",
            "UUID": "a05e85c0-fc4b-38af-9b61-514a3e8cd7e3",
            "start": 1450193684107249704,
            "end": 1450193776505076263,  <<<< incorrect 
            "nbEvents": 241566,
            "path": "/home/user/git/tracecompass-test-traces/ctf/src/main/resources/context-switches/context-switches-kernel",
            "properties": {
                "hostname": "\"qemu1\"",
                "kernel_version": "\"#1 SMP PREEMPT Sat Dec 12 14:52:43 CET 2015\"",
                "tracer_patchlevel": "3",
                "clock_offset": "1450192747804379176",
                "domain": "\"kernel\"",
                "sysname": "\"Linux\"",
                "host ID": "\"40b6dd3a-ffff-431e-92ef-8c4dafe14627\"",
                "kernel_release": "\"4.1.13-WR8.0.0.0_standard\"",
                "tracer_name": "\"lttng-modules\"",
                "clock_scale": "1.0",
                "tracer_major": "2",
                "tracer_minor": "6"
            },
            "indexingStatus": "COMPLETED"
        },
        {
            "name": "context-switches-ust",
            "UUID": "8cf97285-b1cd-321a-b59d-95df7c634dc0",
            "start": 1450193699155527373,
            "end": 1450193747895027378,
            "nbEvents": 3934,
            "path": "/home/user/git/tracecompass-test-traces/ctf/src/main/resources/context-switches/context-switches-ust",
            "properties": {
                "hostname": "\"qemu1\"",
                "clock_offset": "1450192743562703624",
                "domain": "\"ust\"",
                "host ID": "\"40b6dd3a-ffff-431e-92ef-8c4dafe14627\"",
                "tracer_name": "\"lttng-ust\"",
                "clock_scale": "1.0",
                "tracer_major": "2",
                "tracer_minor": "6"
            },
            "indexingStatus": "COMPLETED"
        }
    ]
}
```

### Steps to Reproduce:

1. Open context-switches experiment of the tracecompass-test-traces repo
2. Open the events table
3. Search for the first ust event
4. list the experiment with the python client (tsp_pyton_client --getExperiment <uuid>

Note: That the endtime of the kernel trace could be incorrect instead of the endtime of the ust trace. This depends on the order of the traces being opened and which trace receives the negative offset value of half of the difference of both clock offsets.

### Additional Information

- Operating System: Any
- Trace Compass Version: 11.1

bhufmann:
### What it does

tmf.ctf: Re-initialize start and endtime after transform is applied

With this the start and endtime is correct when opening the experiment that has automatically applied timestamp transformation applied.

fixes #323

### How to test

Follow steps in #323 and verify that start and end times of both traces in the experiment are correct when listing the experiment from the python client.

### Follow-ups

Update `TraceManagerService.instantiateTrace()` to not start indexing of trace during experiment creation. The trace will be indexed when experiment is indexed. 

### Review checklist

### DIFF
--- a/ctf/org.eclipse.tracecompass.tmf.ctf.core/src/org/eclipse/tracecompass/tmf/ctf/core/trace/CtfTmfTrace.java
+++ b/ctf/org.eclipse.tracecompass.tmf.ctf.core/src/org/eclipse/tracecompass/tmf/ctf/core/trace/CtfTmfTrace.java
@@ -218,19 +218,13 @@ public class CtfTmfTrace extends TmfTrace
 
         try {
             this.fTrace = new CTFTrace(path);
-            CtfTmfContext ctx;
             /* Set the start and (current) end times for this trace */
-            ctx = (CtfTmfContext) seekEvent(0L);
-            CtfTmfEvent event = getNext(ctx);
-            if ((ctx.getLocation().equals(CtfIterator.NULL_LOCATION)) || (ctx.getCurrentEvent() == null)) {
-                /* Handle the case where the trace is empty */
-                this.setStartTime(TmfTimestamp.BIG_BANG);
-            } else {
-                final ITmfTimestamp curTime = event.getTimestamp();
-                this.setStartTime(curTime);
-                long endTime = Math.max(curTime.getValue(), fTrace.getCurrentEndTime());
-                this.setEndTime(TmfTimestamp.fromNanos(endTime));
+            CtfTmfContext ctx = setStartAndEndTime(false);
+
+            if (ctx == null) {
+                return;
             }
+
             /*
              * Register every event type. When you call getType, it will register a trace to
              * that type in the TmfEventTypeManager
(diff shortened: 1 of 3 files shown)


MatthewKhouzam (approved):
It works on my machine with a mega-offset

bhufmann:
> This fixes the reported issue.
> 
> I have some hesitation though, because potentially any trace (not just CtfTmfTrace) should consider their start/end time obsolete as soon as a timestamp transform is applied. And it's obsolete regardless of whether or when the indexer needs to be reset.
> 
> I wonder if that could be generalized in TmfTrace with start/end=transformed first event timestamp. It seems maybe the specific implementation in CtfTmfTrace to initialize end time based on first packet end time has not much real benefit as it's only a temporary value that gets quickly updated...
> 
> But we can reconsider that later if it ever becomes an issue.

Acknowledged. Right now only CTF does automatic time offsetting during initialization of the experiment its traces. Other trace types don't do that. Manual offsetting is done outside the trace instances and controlled by the UI model (non-server case). It closes the trace and deletes supplementary folders and when the experiment is re-opened and re-initialized. The automatic offsetting in this PR is a different case. Ideally it should be managed by the classes that instantiates the traces. 

I think this PR is good compromise until we have need for a broader solution.

PR #326, opened 2025-09-25
fix: trace server does not notify the client when the state system is…

aaanh:
… empty and the analysis is completed, resulting in infinite polling on the client side.

### What it does

Adds a condition check on whether or not the analysis is complete and return a sufficient response to the client.

### How to test

- Run trace server (incubator) in debug mode with Eclipse IDE
- Run theia-trace-extension locally
- Load a valid UST trace without any entry or exit calls (i.e. empty)
- Open Flame Chart view
- Should show an infinite analysis running loader

### Follow-ups

n/a

### Review checklist


bhufmann (approved):
Thanks for the fix.

PR #328, opened 2025-10-03
tmf.core: Add event type children to histogram data provider

MatthewKhouzam:
[Added] lines in histogram

Change-Id: I142886fe9fa8e154ef2825a5a01c4f51a48385ea

### What it does

- Add individual event type histogram support to HistogramDataProvider
- Track event type IDs in fEventTypeIds map
- Extend fetchTree() to include event type children from statistics
- Extend fetchXY() to handle event type histogram requests
- Add getEventTypeHistogram() method for per-type density data
- Change EventDensityViewer to use AREA chart style

### How to test

Open a trace. It will have all the event types. This allows users to view histogram density per event type in addition to total and lost events.

note: this commit was made with the assistance of Claude Sonnet 4

### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template


bhufmann:
@MatthewKhouzam There are tests failing that are related to the change. Please have a look.

PR #334, opened 2025-10-10
tmf: Add raw events data provider

PatrickTasse:
### What it does

Add TmfRawEventsDataProvider and its factory. Add them to
o.e.tc.tmf.core.dataprovider extension point.

### How to test

Query the outputs and the data provider with an updated client.

### Follow-ups

- Improve performance of backward read with type and/or field filters.
- Improve robustness against out of memory error when requested count is too large or absent.

### Review checklist


tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/internal/provisional/tmf/core/model/events/TmfRawEventsDataProvider.java
```diff
@@ -0,0 +1,224 @@
+/**********************************************************************
```
MatthewKhouzam: The goal of the raw events provider is to get typed data out fast, please add another option for compare as instead of just equals.

Issue #412, opened 2026-06-05
User Experience: Not seeing pie charts when selecting 1 event type


GuillaumeZhang321:
### Bug Description:
In statistics view, user gets shown no data when selecting a single event type.

Instead of showing nothing, we can enhance the user experience by still showing the global pie chart.

### Steps to Reproduce:

1. Open a trace with more than two event types
2. Select a window of traces that contains only one event type.
3. "No Data" is shown instead of any pie charts

### Fix
In `org.eclipse.tracecompass.internal.tmf.ui.viewers.piecharts.TmfPieChartViewer.refresh`, instead of making an empty viewer when less than two events types are selected, show  the change shows the global pie chart.

### Additional Information

- Operating System: Windows 11
- Trace Compass Version: 11.3

PR #413, opened 2026-06-05
Add fallback in PieChartViewer refresh to not show empty pie chart viewer

GuillaumeZhang321:
… but not enough event types in the selection, show the global pie chart instead.

### What it does
In `org.eclipse.tracecompass.internal.tmf.ui.viewers.piecharts.TmfPieChartViewer.refresh`, instead of making an empty viewer when less than two events types are selected, show  the change shows the global pie chart.

### How to test

In statistics view, select only one event type and see that instead of "No Data" being shown, the global pie chart is left.

### Follow-ups
Addresses #412 

### Review checklist

- [x ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

## Release Notes

* **Bug Fixes**
  * The pie chart viewer now falls back to showing relevant global data (rather than an empty state) when the current selection has too few non-zero event types, improving visibility and avoiding blank charts for small/insufficient selections.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/internal/tmf/ui/viewers/piecharts/TmfPieChartViewer.java
+++ b/tmf/org.eclipse.tracecompass.tmf.ui/src/org/eclipse/tracecompass/internal/tmf/ui/viewers/piecharts/TmfPieChartViewer.java
@@ -422,8 +422,9 @@ public class TmfPieChartViewer extends Composite {
 
                 // Check if the selection is empty or if
                 // there is enough event types to show in the piecharts
+                // if there aren't enough, show the global pie chart
                 if (nbEventsType < 2) {
-                    getCurrentState().newEmptySelection(this);
+                    getCurrentState().newGlobalEntries(this);
                 } else {
                     getCurrentState().newSelection(this);
                 }


MatthewKhouzam (approved):
Looks good to me!

MatthewKhouzam:
Just a friendly reminder, if you used AI in any patch (I doubt this one has AI), you need to bring it up in the commit message. Like "This patch was created with the assistance of chatgpt 5.5"

GuillaumeZhang321:
> Looks good to me!

There seems to be a CI check fail that matches the CI failure on main, is this normal?

Also, would it be possible to include this change in some form of nightly build for trace compass 11.3 since I am working on a project in Eclipse 2026-03? Thanks

MatthewKhouzam:
Looks like this work is fine, but it fails because of the upgrade. rebase it and it should work.

GuillaumeZhang321:
> Looks like this work is fine, but it fails because of the upgrade. rebase it and it should work.

Sorry but I might be confused. Which commit do I rebase on top of? My fork is built on top of the current latest commit on main already.

MatthewKhouzam:
The problem is that the "OK" button got replaced with an "APPLY" button in the tests. not your code, it's the platform that's changing. So that's why I'm saying a rebase would fix it. @arfio could you confirm?

arfio:
The release is today, after that the rebase will fix the issue. For now the build is not using the new platform by default.

GuillaumeZhang321:
> The release is today, after that the rebase will fix the issue. For now the build is not using the new platform by default.

Just making sure, if I want to use an official eclipse download repo link, this PR's changes will only be reflected in the tracecompass 12.0 links? There are no possibilities to have the PR change on an 11.3 version of tracecompass without me using my local version of tracecompass directly?

Thanks!

GuillaumeZhang321:
Hello, 

Would it be possible to restart the workflow check for this PR and #416  ?

Thanks!

GuillaumeZhang321:
Hi @MatthewKhouzam, could you please help me merge this PR unless there is something wrong?

Thank you!

arfio:
> > The release is today, after that the rebase will fix the issue. For now the build is not using the new platform by default.
> 
> Just making sure, if I want to use an official eclipse download repo link, this PR's changes will only be reflected in the tracecompass 12.0 links? There are no possibilities to have the PR change on an 11.3 version of tracecompass without me using my local version of tracecompass directly?
> 
> Thanks!

No this commit will only be included in the 12.1 release which is scheduled for 2026-09. If you want to have a version of Trace Compass including this change you can use the [nightly build](https://download.eclipse.org/tracecompass/master/rcp/) when this patch is merged.

arfio:
Also I am ready to approve this patch, but I do not observe either the issue or the fix on my end, is it possible to include a small gif either highlighting the fix or the bug?

GuillaumeZhang321:
basic_example_issue_pie_chart

Here's an example with the ros2 trace example from [tracecompass test traces](https://eclipse.googlesource.com/tracecompass/tracecompass-test-traces) running on tracecompass 12.0. The issue only seems to arise when the automatic initial selection is of 1 event type only. Such is the case for the ros2 traces. From the debugging I've done, the root problem seems to be as described in #414 . The fallback piechart viewer done in this pull request adds more robustness that ensures the user does not see nothing in the event of the problem described in #414 .

arfio:
I agree that the issue is present, I need a bit more time to check exactly what is happening there.
I took a quick look and it seems that the StatisticsUpdateJob is called multiple times with different arguments. This fix works, I just need a bit more time to make sure there is not another underlying issue here.

GuillaumeZhang321:
Any updates @arfio?

arfio (approved):
Thank you for the fix!

Issue #415, opened 2026-06-05
Better Dependent Analyses support


GuillaumeZhang321:
### Feature Description: Better Dependent Analyses support 

Hello,

While developing custom analyses, I found the need to make them dependent on each other. Looking online and in the linux kernel cpu analysis provided by this project, I've seen usage of the form 
`
TmfTraceUtils.getAnalysisModuleOfClass(
            getTrace(),
            MyDependency.class,
            MyDependency.ID);
`  inside an overriden `getDependentAnalyses()`. [Example here](https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/master/analysis/org.eclipse.tracecompass.analysis.os.linux.core/src/org/eclipse/tracecompass/analysis/os/linux/core/cpuusage/KernelCpuUsageAnalysis.java#L125)

However, looking into the source code, this strategy is not useful at all for initial load of a trace with automatic analyses that have dependencies. This call leads to a chain of methods culminating in accessing a map in this file through the following getter: [at line 337](https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/master/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/trace/TmfTrace.java#L337).

Looking at the order the analyses are run, we can see that the variable the getter accesses is not filled until after all automatic analyses are ran. [at line 878](https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/master/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/trace/TmfTrace.java#L878). In that way, the getter is only useful after all the analyses are made, which defeats the purpose of using it to look for dependent analyses when opening a trace for the first time. 

I would like to know whether I am managing these dependent analyses in a improper manner or if there is a better way to deal with them. For now, I've instead relied on ordering my plugin.xml to start the analyses in a specified order that satisfies the dependencies but that does not seem robust.

I've also linked a Pull Request which adds a minimal change to the `TmfTrace` class in `org.eclipse.tracecompass.tmf.core.trace.TmfTrace` . In it, simply loop over all discovered available analyses from the source files and add each of them to the Map that is ultimately used when calling 
`
TmfTraceUtils.getAnalysisModuleOfClass(
            getTrace(),
            MyDependency.class,
            MyDependency.ID);
` 
[...]

PR #416, opened 2026-06-05
Add all discovered analyses to map used in getter function used by getDependentAnalyses

GuillaumeZhang321:
### What it does

Addresses #415 

### How to test

* Create numerous automatic modules and set them in any order in the plugin.xml
* Inside of any analysis, make one depend on the others
* Observe, with a step debugger, that the dependent analyses are acquired and scheduled before the analysis that depends on them runs

### Review checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of dependent analysis modules during initialization to ensure proper module registration and management.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/trace/TmfTrace.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/trace/TmfTrace.java
@@ -850,6 +850,19 @@ public abstract class TmfTrace extends TmfEventProvider implements ITmfTrace, IT
         Set<String> oldAnalysisModulesKeys = new HashSet<>(previousAnalysisModules.keySet());
         Set<String> keys = new HashSet<>(newAnalysisModules.keySet());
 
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
You don't allow the same analysis to come in twice, this is good!

MatthewKhouzam:
Was this code AI assisted, if so please add a line in the commit message.

GuillaumeZhang321:
> Was this code AI assisted, if so please add a line in the commit message.

I haven't used AI, should I ask it to verify the changes and then change the commit message to include?

MatthewKhouzam:
No need to change anything then, you're all good!

MatthewKhouzam:
I am expecting it to fail at the OK / APPLY place too.

GuillaumeZhang321:
Hi @MatthewKhouzam, could you please help me merge this PR unless there is something wrong?

Thank you!

arfio (approved):
Thank you for the patch! This seems to fix having automatic analyses dependent on other automatic analyses. At some point, I think refactoring the way these automatic analyses work would be a good idea!

PR #417, opened 2026-06-09
Public API getProviders() in the data provider class

tuanvtdeptre199:
### What it does
Recently, I have wanted to improve my 3rd trace views to work with multiple trace packages from the different trace tools. So I want to public the API getProviders() from class TmfTreeCompositeDataProvider.java and getTrace() from class AbstractTmfTraceDataProvider.java

### How to test
It does not change any logic code, just public some API, so all views in Trace Compass are expected to work normally.

### Review checklist

## Summary by CodeRabbit

* **Refactor**
  * Expanded the core data provider public API to allow retrieving the underlying trace instance directly.
  * Updated tree composite providers to expose their provider list via a public method while returning an unmodifiable view to help prevent external modification.

### DIFF
--- a/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/tree/TmfTreeCompositeDataProvider.java
+++ b/tmf/org.eclipse.tracecompass.tmf.core/src/org/eclipse/tracecompass/tmf/core/model/tree/TmfTreeCompositeDataProvider.java
@@ -188,8 +188,8 @@ public class TmfTreeCompositeDataProvider<M extends ITmfTreeDataModel, P extends
      *
      * @return the list of encapsulated providers
      */
-    protected List<P> getProviders() {
-        return fProviders;
+    public List<P> getProviders() {
+        return Collections.unmodifiableList(fProviders);
     }
 
     /**


arfio:
Can you please rebase your change so that your patch can run the CI completely?

arfio:
Can you please explain in more detail why do you need to access the trace directly to manage multiple traces with one view? 
In my mind, either you need to run an analysis and data provider with an ensemble of trace and you need to use an experiment. Or you have multiple analyses and you need to merge multiple data sources, and you don't need to access the trace only the different dataproviders.

tuanvtdeptre199:
Hi @arfio, I have rebased my change.
Also, for your query, currently I am running an analysis/data provider with an ensemble of traces, and I need to use an experiment.
I think it is better if we can access the trace in the data provider. Because if I or the 3rd side customizes the new TMF trace that extends from the ITmfTrace, they may want to get their trace to do something (For example, as me, I want to get the name of the input trace).
Please help me consider my idea again. If we have a special reason to hide the TMF trace in the data provider from outside, then it is OK for me to use only the child providers.

arfio:
The main reason I am refraining to modify the Data provider API is that it is linked to the TSP. It was done that way so that every analyses that works on Trace Compass standalone would work with the theia trace extension. If you need data about the trace, the data provider would be able to supply it through the already existing API.

tuanvtdeptre199:
@arfio, Thank you for your comment. I understood. If it affects the other extension, then I will revert it. 
But can we publish the API to get the list of providers in the tree composite data provider?

arfio:
To be clearer, the data provider class has been put in to separate the analyses side of things which include the traces from the view side of things. This change has been done to be able to create a server which has a dedicated API in this case the [TSP (trace server protocol)](https://github.com/eclipse-cdt-cloud/trace-server-protocol). The trace extension that uses this protocol has been implemented as a plugin both on [Theia](https://github.com/eclipse-cdt-cloud/theia-trace-extension) and [Vscode](https://github.com/eclipse-cdt-cloud/vscode-trace-extension).

The data provider is exposed through the [DataProviderService](https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/blob/master/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java) in the incubator trace server implementation.
All of the public getters are implemented by different types of data providers.

If getProviders() were public, then it will not be possible to replicate your view in the extension and as it is made public, other people might use it and it will separate the users in two groups.

I will be more than happy to help you implement the feature that you are trying to make with the current API, and I am sure it will improve Trace Compass. Is it possible for you to explain in more details the issue you are currently facing so that we may help you get it working?

tuanvtdeptre199:
@arfio , Here is my problem (I'm not sure if I've expressed it clearly, so please forgive me if you're confused.) 
I created the new trace view -> and I created the Data provider factory, which implements the class `IDataProviderFactory` -> I override the method `public ITmfDataProvider createDataProvider(ITmfTrace trace)`

- In the single trace package, it is simple, I just check if the input `ITmfTrace` is an instance of my extended trace class -> If yes, I will return my extended data provider class. 
- When deals with experiment trace, that have more than one trace, I choose to return `new TmfTreeXYCompositeDataProvider<>(providers, myproviderTittle,
				myproviderID);` which `providers` is the list of my data provider object instances that extends from `ITmfTreeXYDataProvider<@NonNull TmfTreeDataModel>`, each provider corresponding to 1 child trace (that is an instance of my extended trace class) in the experiment trace. I think it is reasonable, because the experiment is an esemmble of the traces.
- It still works ok, but the problem occur when I want to draw more addtional tooltip information into my chart -> I have created my API to get this information in dataprovider class -> Now, in the tooltip provider class/ chart viewer class,I want to get the list of my data provider class that contains in the ojbect  `TmfTreeXYCompositeDataProvider` of the experiments trace. (To more details, at this time the user open the experiments and trace views, the method `public void loadTrace(ITmfTrace trace)` at chart viewer class is hit -> In this API I can get object `TmfTreeXYCompositeDataProvider` of the experiment trace)
- However, as I can not access the list of my provider class that I stored in the object `TmfTreeXYCompositeDataProvider` before, So I can not add my additional information into the chart view. This is my problem.
- Of course, I can work around to do it. But I think the best thing is public the API 'getProviders' (getTrace() also, but it still OK if we can not public it). 
- Could you please help me check my problem? Thank you in advance

arfio:
So I looked into it, and the tooltip interface would be a perfect fit if it were implemented for the XY dataproviders.
The tooltip is implemented for timegraphs right now. I will be putting up a small patch to introduce tooltips for xy dataproviders. Using this interface you would be able to change your view to query the dataprovider for a tooltip and the dataprovider would have access to the information.

Would that be enough to implement your 3rd party view ? @tuanvtdeptre199

tuanvtdeptre199:
Hi @arfio , Thank you for your support!. Could you please share me details about your changes? I want to check it to consider whether it can work for my case or not.

arfio:
The input parameters are written in DataProviderService:524-537. The output would be the same as the TimeGraphTooltip so you can look at an example implementation in FlameChartDataProvider:273-320.

My change is very small right now as I did only part of what is needed: https://github.com/arfio/org.eclipse.tracecompass/commit/9aa503d02230743645046fde85f4f47672237190

tuanvtdeptre199:
Hi @arfio, so you want to support fetch tooltip in the XY data provider, just like we do with the Time Graph data provider? I have also created a new view that implements the `ITimeGraphDataProvider` and the `fetchTooltip` method. So, I think it is a great idea for the XY data provider.
Please notify me again when you complete this improvement. Thank you for your support!

arfio:
Hi @tuanvtdeptre199, I do not have enough time currently to make this change. I pointed toward a small modification to point where you should begin to add this feature. I would be happy to guide you around the codebase and answer questions.

jonahgraham:
@arfio thank you for your support of @tuanvtdeptre199 this far.

@tuanvtdeptre199 I believe you have all the information needed to proceed on this task now. Please feel free to ask follow up questions and  @arfio or I will attempt to help and guide you to the final solution.

tuanvtdeptre199:
Hi @arfio , @jonahgraham 
Sorry for my slow response. I understand your idea. I will implement it and get back to you if I have any further questions. Thank you!

PR #422, opened 2026-07-02
xy: Add mipmap-based downsampling for XY data providers

MatthewKhouzam:
### What it does
Introduce max-per-bucket aggregation backed by pre-computed mipmap levels in the state system. This replaces trivial point-sampling with spike-preserving downsampling that guarantees visual correctness at every zoom level.

New components:
- AbstractXYStateProvider (provisional API): base state provider with ergonomic mipmap helpers (modifyMipmapAttribute, incrementMipmap*)
- TimeMipmapFeature: time-based mipmap with power-of-10 levels (10ns, 100ns, 1us, 10us, ...) aligned to natural time units
- MipmapXYQueryHelper: internal utility that auto-detects mipmap sub-attributes and replaces Y values with queryRangeMax per bucket

Integration:
- AbstractTreeCommonXDataProvider.fetchXY() auto-enhances results
- AbstractTreeGenericXYCommonXDataProvider.fetchXY() same (timestamps only)
- AbstractTreeDataProvider exposes getIdToQuark() for ID-to-quark mapping

Migration (proof of concept):
- CounterStateProvider now extends AbstractXYStateProvider
- Uses mipmap helpers instead of StateSystemBuilderUtils
- State system version bumped to trigger rebuild

### How to test
Open some xy plots.

### Follow-ups
Make all XY plots use delta based storage.

### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template


MatthewKhouzam:
@arfio, @JonatanAntoni fyi

JonatanAntoni:
@thorstendb-ARM

JonatanAntoni:
The issue we are looking to solve can be reproduce with e.g. with a sampled sine wave like this:

Bildschirmfoto 2026-07-07 um 14 37 18

Zooming out, so that one or more periods of the sine wave collapse onto a single pixle the current downsampling causes display effects like this:

Bildschirmfoto 2026-07-07 um 14 37 07

Instead, something like the following would be better for us:
Bildschirmfoto 2026-07-07 um 14 53 41

MatthewKhouzam:
Ah, this makes more sense. it is an interesting problem... Basically how would you like to see it solved? we can make an "envelop" per pixel. I usually look at an oscilloscope as a reference, and what is shown above is how it would display it.

JonatanAntoni:
Here is how I solved this in my TypeScript TSP implementation to get to the latest picture:
1. Bucketize the whole sample set to fixed time boundaries that don't change when panning the view.
2. For each bucket that is (partially) within the selected view range, select four values: first, min, max, and last. 

Ultimately, this leads to a factor 4 oversampling. I.e., if the client requests 1000 samples, this method actually sends back 4000 samples. The current graph implementation then results drawing 4 samples onto the same horizontal position. Of course, this is wasting some bandwidth and there might be a more elegant solution as well.

MatthewKhouzam:
Now oversampling works too. The mipmap and oversampling offer two solutions to the problem. What you're proposing could still miss a spike, but it's less likely. Let's discuss/nerd out on wednesday's meeting.

MatthewKhouzam:
Here is an illustration of all three strategies as per my understanding.

As-is
sampling-aliasing
With oversampling, much better.
oversampling
With mipmap, it's garanteed not to lose the max
mipmapping

thorstendb-ARM:
Max-per-bucket preserves positive spikes, but it does not make waveform rendering visually correct.

For waveforms, a bucket needs to retain the envelope and continuity information: min, max, and the exit/last value used to connect to the next bucket. Without that, aliasing is still possible because the renderer may connect representative points in a way that misrepresents the original curve.

I attached a graphic showing an expanded view of compressed buckets.
image

JonatanAntoni:
@MatthewKhouzam, my approach isn't actually oversampling as I do not pick equally spaced samples from the original signal. We are picking the highest and lowest Y-values from a bucket regardless of their X-values and put them together on the same scaled X-value. Its similar to the mipmap-approach but with more values so that positive and negative extrema are preserved.

PR #427, opened 2026-08-05
ctf: fix bits interpreted as bytes in index creation

arfio:
### What it does

Fix how packet contexts are mapped into memory. The packet context size was in bits not bytes, a division was added to fix that issue.

### How to test

Use the following trace:  [the-trace.zip](https://github.com/user-attachments/files/30747066/the-trace.zip)

### Follow-ups

N/A

### Review checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed packet context indexing to ensure trace data is mapped using the correct byte-based size.
  * Improved reliability when reading and navigating CTF traces.


MatthewKhouzam (approved):
oops!

